PR #2, opened 2020-04-22
add first draft for SectionResponse with Stages

harryurban:
as roughly discussed here (desktop viewport), see second comment with enhanced model with config

Screenshot 2020-04-22 at 15 46 34

model shown in this image has no configs, only articles

### DIFF
--- a/proto/section_response.proto
+++ b/proto/section_response.proto
@@ -1,0 +1,15 @@
+syntax = "proto3";
+
+import "content.proto";
+
+option java_package = "de.stroeer.model.wrapper.generated";
+option java_multiple_files = true;
+
+message SectionResponse {
+  repeated Stage stages = 1;
+}
+
+message Stage {
+  repeated Article main = 1;
+  repeated Article sib = 2;
+}
--- a/proto/section_response.proto
+++ b/proto/section_response.proto
@@ -10,6 +10,16 @@ message SectionResponse {
 }
 
 message Stage {
-  repeated Article main = 1;
-  repeated Article sib = 2;
+  ArticleGroup main = 1;
+  ArticleGroup sib = 2;
+}
+
+message ArticleGroup {
+  repeated Article articles = 1;
+  GroupConfig config = 2;
+}
+
+message GroupConfig {
+  string label = 1;
+  string href = 2;
 }


harryurban:
Visualize commit `add config for linkable label above a stage/group`
SectionResponse

proto/section_response.proto
```diff
@@ -0,0 +1,25 @@
...
+message SectionResponse {
+  repeated Stage stages = 1;
+}
+
+message Stage {
+  ArticleGroup main = 1;
+  ArticleGroup sib = 2;
+}
+
+message ArticleGroup {
```
bobaaaaa, line 17: IMO this is a `Stage`

proto/section_response.proto
```diff
@@ -0,0 +1,25 @@
...
+import "content.proto";
+
+option java_package = "de.stroeer.model.wrapper.generated";
+option java_multiple_files = true;
+
+message SectionResponse {
+  repeated Stage stages = 1;
+}
+
+message Stage {
```
bobaaaaa, line 12: `Section` maybe better?

proto/section_response.proto
```diff
@@ -0,0 +1,25 @@
...
+  ArticleGroup main = 1;
+  ArticleGroup sib = 2;
+}
+
+message ArticleGroup {
+  repeated Article articles = 1;
+  GroupConfig config = 2;
+}
+
+message GroupConfig {
```
bobaaaaa, line 22: Actually this is a `StageConfig`

proto/section_response.proto
bobaaaaa, line 17: I don't like this name, but I don't have a better one. What happened to the idea to introduce a `Teaser` model on `Section`'s?

proto/section_response.proto
```diff
@@ -0,0 +1,25 @@
...
+option java_package = "de.stroeer.model.wrapper.generated";
+option java_multiple_files = true;
+
+message SectionResponse {
+  repeated Stage stages = 1;
+}
+
+message Stage {
+  ArticleGroup main = 1;
+  ArticleGroup sib = 2;
```
bobaaaaa, line 14:
```suggestion
  repeated ArticleGroup groups = 1;
```

Does this work too?

proto/section_response.proto
bobaaaaa, line 12: IMO: The whole meaning of a `Stage` is now lost. Maybe we reverse the naming a bit:

```proto
message Group {
  Stage main = 1
}

message Stage {
  repeated Teaser content = 1
  StageConfig config = 2
}

message StageConfig {
 string label = 1
}
```

harryurban:
thanks for the feedback. lets focus on #4

PR #21, opened 2020-05-20
generate go code

moritzzimmer:
`make LANGUAGE=go OUTPUT=./go`

no ci or module management so far


shobhitsharma:
@moritzzimmer Only gitHub-action is missing from workflow.

PR #22, opened 2020-05-25
draft for Article BFF

moritzzimmer:
PR to discuss the article (page) BFF used by paper (and future AMP) service to get relevant data.

- moved all paper related api to new folder `web`
- introduced separate `WebArticle` message
- moved core article grpc service next to proto message (needs a bit of work in polyphase)
- tried to used meaningful file names


stroeer/article/v1/service.proto
```diff
@@ -10,9 +10,34 @@ option java_package = "de.stroeer.article.v1";
...
+  // rendering (canonical web or AMP) an Article with the given id.
+  rpc GetArticlePage (GetArticlePageRequest) returns (ArticlePage) {}
 }
 
 message GetArticleRequest {
   string id = 1;
 }
 
+message GetArticlePageRequest {
+  string id = 1;
```
thisismana, line 24: what's the `id` in this case? A path like `/` or `/culture/`?

stroeer/article/v1/service.proto
```diff
@@ -10,9 +10,34 @@ option java_package = "de.stroeer.article.v1";
...
+}
+
+message RelatedArticle {
+  // we could also use an stroeer.core.v1.Article here but maybe it's worth to have a Teaser abstraction
+  // including only teaser relevant attributes?!
+  stroeer.core.v1.Teaser teaser = 1;
+  RelatedArticleType role = 2;
+}
+
+enum RelatedArticleType {
```
thisismana, line 39: `RelatedArticleRole` ?

stroeer/article/v1/service.proto
```diff
@@ -10,9 +10,34 @@ option java_package = "de.stroeer.article.v1";
...
+  RELATED_ARTICLE_TYPE_UNSPECIFIED = 0;
```
thisismana, line 40: Shouldn't we just filter out relations unknown by our code during transformation?

stroeer/article/v1/service.proto
moritzzimmer: the CMS id

stroeer/article/v1/service.proto
moritzzimmer: protobuf lint rule to always include unknown at `0`

stroeer/web/section/v1/web_section_service.proto
```diff
@@ -0,0 +1,21 @@
...
+import "stroeer/web/article/v1/web_article.proto";
+
+option java_multiple_files = true;
+option java_package = "de.stroeer.web.section.v1";
+
+service WebSectionService {
+  rpc GetSectionPage (GetSectionPageRequest) returns (GetSectionPageResponse) {}
+}
+
+message GetSectionPageRequest {
```
harryurban, line 14: for now, I guess?
Will need to change, as soon as we have stages, with labels, and configs for the page.

PR #23, opened 2020-05-25
Build in Dockerfile that works for java and node

harryurban:
- Add Dockerfile that works for node and java
- Use docker image from Dockerfile for make
- fix paths in Makefile for always running in docker

TODO:
- publish docker image on hub.docker.com e.g. to use it
- clean package.json from obsolete dependencies

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -57,11 +57,8 @@ jobs:
           java-version: 11
           java-package: jdk
 
-      - name: setup protoc
-        run: ./install_proto.sh
-
       - name: make
-        run: make PROTOC=./bin/protoc LANGUAGE=java OUTPUT=./java/src/main/java
+        run: make
 
       - name: build
         run: ./gradlew clean build
@@ -80,11 +77,8 @@ jobs:
         run: npm install
         working-directory: node
 
-      - name: setup protoc
-        run: ./install_proto.sh
-
       - name: make
-        run: make PROTOC=./bin/protoc LANGUAGE=node OUTPUT=./node
+        run: make LANGUAGE=node
 
       - name: Run node checks
         run: npm test
--- a/Makefile
+++ b/Makefile
@@ -3,40 +3,41 @@
 # language. However, it does not compile the generated code into final
 # libraries that can be directly used with application code.
 #
-# Supported (tested) languages: java, node (TypeScript), go
+# Supported (tested) languages: java, node (TypeScript)
 #
 # Syntax example:
-#		make LANGUAGE=node OUTPUT=./node
+#		make LANGUAGE=node
 #
 
 DIR = $(shell pwd)
 JAVA_DIR = ./java/src/main/java
 NODE_DIR = ./node
-GO_DIR = ./go
 
 OUTPUT ?= $(JAVA_DIR)
 LANGUAGE ?= java
 GRPCPLUGIN ?= /usr/bin/protoc-gen-grpc-$(LANGUAGE)
-# with https://github.com/TheThingsIndustries/docker-protobuf/issues/9 we could use this image also for TypeScript
-PROTOC ?= docker run --rm -v $(DIR):$(DIR) -w $(DIR) thethingsindustries/protoc
+PROTOC ?= docker run --rm -v $(DIR):$(DIR) -w $(DIR) stroeer/protoc-dockerized
 
-FLAGS+= --proto_path=.
+FLAGS+= --proto_path=$(DIR)
 ifeq ($(LANGUAGE),node)
-FLAGS+= --plugin=protoc-gen-ts=$(NODE_DIR)/node_modules/.bin/protoc-gen-ts
-FLAGS+= --plugin=protoc-gen-grpc=$(NODE_DIR)/node_modules/.bin/grpc_tools_node_protoc_plugin
-FLAGS+= --js_out=import_style=commonjs,binary:$(NODE_DIR)
-FLAGS+= --ts_out=service=grpc-node:$(NODE_DIR)
-FLAGS+= --grpc_out=$(OUTPUT)
-else ifeq ($(LANGUAGE),go)
-FLAGS+= --$(LANGUAGE)_out=plugins=grpc:$(OUTPUT)
+	# make node OUTPUT overridable, but with NODE_DIR as default
+	ifeq ($(OUTPUT), $(JAVA_DIR))
+		OUTPUT = $(NODE_DIR)
+	endif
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -18,12 +18,12 @@ OUTPUT ?= $(JAVA_DIR)
 LANGUAGE ?= java
 GRPCPLUGIN ?= /usr/bin/protoc-gen-grpc-$(LANGUAGE)
 # with https://github.com/TheThingsIndustries/docker-protobuf/issues/9 we could use this image also for TypeScript
-PROTOC ?= docker run --rm -v $(DIR):$(DIR) -w $(DIR) thethingsindustries/protoc
+PROTOC ?= docker run --rm -v $(DIR):$(DIR) -w $(DIR) protoc-harry
```
harryurban, line 21: of course this is temporary until image is pushed and available

Makefile
```diff
@@ -18,12 +18,12 @@ OUTPUT ?= $(JAVA_DIR)
...
 
-FLAGS+= --proto_path=.
+FLAGS+= --proto_path=$(DIR)
 ifeq ($(LANGUAGE),node)
-FLAGS+= --plugin=protoc-gen-ts=$(NODE_DIR)/node_modules/.bin/protoc-gen-ts
-FLAGS+= --plugin=protoc-gen-grpc=$(NODE_DIR)/node_modules/.bin/grpc_tools_node_protoc_plugin
+FLAGS+= --plugin=protoc-gen-ts=/node_modules/.bin/protoc-gen-ts
```
harryurban, line 25: paths are absolut cause we know the location in the image

Makefile
```diff
@@ -36,7 +36,7 @@ FLAGS+=	--plugin=protoc-gen-grpc=$(GRPCPLUGIN)
 FLAGS+= --grpc_out=$(OUTPUT)
 endif
 
-all: clean stroeer/*
+all: stroeer/*
```
harryurban, line 39: removed clean from `all` target, because running `java` and `node` sequentially deletes the generated files from the previous run

Makefile
```diff
@@ -51,8 +51,8 @@ clean: ## Deletes all generated files
 	@echo "+ $@"
 	rm -rf $(JAVA_DIR)  || true
 	rm -rf $(GO_DIR)  || true
-	rm -R `find node -type d \( -iname "*" ! -iname "node_modules" ! -iname "tests" \) -mindepth 1 -maxdepth 1` || true
+	rm -rf `find $(NODE_DIR) -type d \( -iname "*" ! -iname "node_modules" ! -iname "tests" \) -mindepth 1 -maxdepth 1` 2> /dev/null || true
```
harryurban, line 54: get rid of usage printout of `rm` when no files need to be deleted by passing the errout to `/dev/null`

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM node:lts-stretch
+ARG JAVA_LIB_VERSION=1.29.0
+ARG PROTOC_VERSION=3.11.4
+RUN npm install ts-protoc-gen grpc-tools grpc
```
bobaaaaa, line 7:
```suggestion
```

Lets use the local ones in `./node/package.json`?

bobaaaaa, line 7: Or

```suggestion
RUN npm install --gobal ts-protoc-gen grpc-tools
```

Falls das hier geht:
```
ENV PATH="(npm bin ts-protoc-gen --global):${PATH}"
ENV PATH="(npm bin grpc-tools --global):${PATH}"
```

harryurban: not working, agreed with @bobaaaaa to leave it as it is

harryurban: agreed with @bobaaaaa to instead remove these depds from package.json and leave the version mngmt in `build_docker.sh`

Makefile
moritzzimmer: if we keep on specifying distinct output directories (`OUTPUT=./node`) cleaning should be fine?!

Makefile
moritzzimmer: `FLAGS+= --grpc_out=$(OUTPUT)` should be probably be `FLAGS+= --grpc_out=$(NODE_DIR)`

Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM node:lts-stretch
```
moritzzimmer, line 1: this results in a very large image (1.23GB), we might leverage multi-stage docker builds as described in https://github.com/nodejs/docker-node/issues/282

.github/workflows/pull-request.yml
```diff
@@ -52,17 +52,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: make
+        run: make LANGUAGE=java
```
moritzzimmer, line 56: still the default 👍  (so could be just `make`)

Makefile
moritzzimmer: if we fix the `grpc_out` for `node` cleaning should be fine

harryurban: but the clean task will remove java files if we run the node generation afterwards or vice versa

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# Download Java Lib
+RUN wget https://repo1.maven.org/maven2/io/grpc/protoc-gen-grpc-java/$JAVA_LIB_VERSION/protoc-gen-grpc-java-$JAVA_LIB_VERSION-linux-x86_64.exe -O /installer/protoc-gen-grpc-java
+
+# protoc
+RUN wget https://github.com/protocolbuffers/protobuf/releases/download/v$PROTOC_VERSION/protoc-$PROTOC_VERSION-linux-x86_64.zip -O protoc.zip
+RUN mkdir -p /installer/protoc
+RUN unzip -o protoc.zip -d /installer/protoc/
+
+FROM node:lts-stretch-slim
```
bobaaaaa, line 22:
```suggestion
FROM node:lts-alpine
```

Das dürfte noch kleiner sein.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+USER root
+COPY --from=installer /node_modules /node_modules
+COPY --from=installer /installer/protoc /usr/bin/protoc
+COPY --from=installer /installer/protoc-gen-grpc-java /usr/bin/
```
bobaaaaa, line 26:
```suggestion
USER node
COPY --chown=node:node --from=installer /node_modules /node_modules
COPY --chown=node:node --from=installer /installer/protoc /usr/bin/protoc
COPY --chown=node:node --from=installer /installer/protoc-gen-grpc-java /usr/bin/
```

Als `root` ist nicht so gut, oder?

harryurban:
merging with root user because with node user we got permission denied during CI.
merging without alpine because of permission problems during docker build

PR #30, opened 2020-06-10
Change GetWebSectionPageResponse to add teaser headlines

harryurban:
Change GetWebSectionPageResponse to meet requirements for BUZZ-169 (Get headlines for the Schlagzeilenbox)


harryurban:
My proposal is to mutate the WebArticle for the SectionPageResponse to exchange the headline with the short headline, if present (see Schlagzeilen)

PR #46, opened 2020-09-17
chore: version bump + @grpc/grpc-js

bobaaaaa:
## What

- bumped `protoc` + all plugins
- (node) replaced `grpc` with `@grpc/grpc-js`
- re-generated `go` files (is this ok?)

## Why

- housekeeping
- `@grpc/grpc-js` is the new reference impl. It is 100% javascript instead of some C binaries

## Info

For nodejs this is a **BREAKING CHANGE**. I tested this version locally with paper.


bobaaaaa:
@harryurban How do we release a new docker image? Should I split this PR: first only the version bump, second the `@grpc/grpc-js` replacement?

moritzzimmer (commented):
LGTM, shall we update the docker image @bobaaaaa ?

bobaaaaa:
@moritzzimmer yes please. After that the node test will be green again.

moritzzimmer (approved):

PR #49, opened 2020-09-25
feature / docker release action

vmk1vmk:
## What 

Added a Github action workflow to build and publish the `protoc` Docker image when changes to it occurred.

## How

feature: added Github action to release internally used Docker image on push to master to GHCR.

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -1,0 +1,27 @@
+name: docker-release
+
+on:
+  push:
+    branches:
+      - "master"
+    paths:
+      - "Dockerfile"
+      - "build_docker.sh"
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Docker build
+        run: ./build_docker.sh
+
+      - name: GHCR login
+        run: docker login ghcr.io -u ${{ github.repository_owner }} --password ${{ github.token }}
+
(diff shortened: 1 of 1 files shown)


moritzzimmer:
@vmk1vmk please rebase, otherwise 👍

moritzzimmer (changes requested):
we shouldn't use `latest` but rather build and use images with a specific version, e.g. use the `protoc` version as a tag

harryurban:
> we shouldn't use `latest` but rather build and use images with a specific version, e.g. use the `protoc` version as a tag

but maybe in another PR?
currently we also only tag and use latest. in my opinion this is already an improvement to the current situation regarding automation.
your change request is an improvement on top of that, regarding reproducability.

harryurban (approved):

moritzzimmer:
> > we shouldn't use `latest` but rather build and use images with a specific version, e.g. use the `protoc` version as a tag
> 
> but maybe in another PR?
> currently we also only tag and use latest. in my opinion this is already an improvement to the current situation regarding automation.
> your change request is an improvement on top of that, regarding reproducability.

we ran into issues while upgrading `grpc_js`, so we moved away from using latest @harryurban , see https://github.com/stroeer/tapir/commit/e7caa19aac8bd39c597fd4606624da2faf372f28

moritzzimmer (approved):
the hard-coded docker tag should be aligned with a strategy of tagging and releasing new image versions

PR #56, opened 2020-09-30
Mandatory root fields and strongly typed dates

### LINKED ISSUE #32: Element relations as enum
Currently only the `relation` field of `Element` is not an `enum`. When things about the article body are clear we should migrate here to an `enum` as well.

```proto
message Element {
  ElementType type = 1;
  string guid = 2;
  repeated string relations = 3;
  repeated Asset assets = 4;
  string article_id = 5;
}
```

### LINKED ISSUE #41: section path on GetWebSectionPageResponse
To generate **breadcrumbs** we need the `section_path` on the `GetWebSectionPageResponse` message.

Currently a `WebArticle` has a `section_path` field. Maybe it make more sense if both `GetWebSectionPageResponse` and `GetWebArticlePageResponse` includes a more structured `section_path` field.

#### Proposal

```protobuf
message GetWebSectionPageResponse {
  // ...
  repeated string section_path = ?
}
```

```ts
const sectionPath = ['', 'sport', 'fussball'];
```

### LINKED ISSUE #45: data type of article ids
we currently use `string` as the data type for article ids (in `core` and `web`). If we keep on using numeric database ids, we should migrate this to something more efficient like `uint64`.

thisismana:
* introduced new mandatory top level fields `top_line` and `summary` which are similar to `headline`. We should either move all three into `fields` or move all three into the root. Front end decided against the `field` approach, because "there have to be fallbacks" (`.getOrElse(???)`)

* Added strongly editorial dates as strongly typed fields to `Metadata`, currently they reside in `fields` as ISO-formatted Strings which is also not "strongly typed"
* Added Metadata to WebArticle to give access to the article's strongly typed dates: `updated`, `created`, `last_edited`, `last_published` and also the technical `last_modification`
* numeric ids (fixes #45)
* structured section data (fixes #41)
* element relation as enum (fixes #32)


stroeer/web/article/v1/web_article.proto
```diff
@@ -19,6 +19,7 @@ message WebArticle {
   repeated stroeer.core.v1.Element elements = 8;
   repeated stroeer.core.v1.Author authors = 9;
   repeated stroeer.core.v1.BodyNode body_nodes = 10;
+  stroeer.core.v1.Metadata metadata = 11;
```
bobaaaaa, line 22: With this change: why do we need `WebArticle` anymore? I guess its now 100% the same to `Article`. From a FE view this metadata object feels weird to me.

thisismana: For a start we'll add the detailed author-entities later on, which is not part of `Article`. There may be more to come, I fear.

Otherwise, where do you suggest putting those 4 dates? Protobuf leverages strongly typed APIs, so `fields.put('key', date.toString()` is not the right approach, imho. Should we could add another Container for dates instead?

harryurban: why do we need the Metadata in the WebArticle? I suppose its sufficient to have these in the Article for paperboy to respond with 404 or to filter unpulished elements?

stroeer/pages/article/v1/article.proto
```diff
@@ -0,0 +1,19 @@
...
+option java_multiple_files = true;
+option java_package = "de.stroeer.pages.article.v1";
+
+message RelatedArticle {
+  stroeer.core.v1.Article article = 1;
+  RelatedArticleRole role = 2;
+}
+
+enum RelatedArticleRole {
+  RELATED_WEB_ARTICLE_ROLE_UNSPECIFIED = 0;
```
moritzzimmer, line 17: rm `WEB`

stroeer/core/v1/article.proto
```diff
@@ -9,47 +9,72 @@ option java_multiple_files = true;
...
+  // profile picture
+}
+
+message Section {
+  string path = 1;
+  string label = 2;
+  Section parent = 3;
+}
+
+message ArticleReferences {
```
moritzzimmer, line 71: only `References`?

stroeer/core/v1/article.proto
```diff
@@ -9,47 +9,72 @@ option java_multiple_files = true;
...
-  string first_name = 1;
-  string surname = 2;
-  string web_url = 3;
+  uint64 id = 1;
+  string first_name = 2;
+  string last_name = 3;
+  // alias
+  // vg-wort id
+  // profile picture
+}
```
bobaaaaa, line 63: We already know that we need the `role`, `web_url`/`path` and some kind of image `element`. Next iteration?

moritzzimmer:
@thisismana guess we need to terraform the `articles` table to alter the key schema

stroeer/web/article/v1/web_article.proto
bobaaaaa, line 22: revert: not needed anymore.

stroeer/core/v1/article.proto
```diff
@@ -9,47 +9,72 @@ option java_multiple_files = true;
...
-  string web_url = 2;
-  string canonical_url = 3;
-  ContentType type = 4;
-  string section_path = 5;
-  string headline = 6;
-  map<string, string> fields = 7;
-  repeated string keywords = 8;
-  Metadata metadata = 9;
-  repeated Element elements = 10;
+  uint64 id = 1;
```
moritzzimmer, line 12: google says https://linter.aip.dev/141/forbidden-types

stroeer/core/v1/article.proto
```diff
@@ -9,47 +9,72 @@ option java_multiple_files = true;
...
+  repeated ElementRelation relations = 2;
+  repeated Asset assets = 3;
 }
 
 message Asset {
   AssetType type = 1;
-  int32 index = 2;
-  map<string, string> fields = 4;
-  Metadata metadata = 5;
+  uint32 index = 2;
```
bobaaaaa, line 51: why `uint32` here and article id is `uint64`?

thisismana: this is a 0-based index, where as the article id is the database id which is somehwere 100,000,000+
In Java it would be the decision between `int` and `long`

stroeer/core/v1/article.proto
thisismana: jira: BUZZ-386 — mozi has to do it.

moritzzimmer: future PR

stroeer/core/v1/article.proto
thisismana: dat java :/

bobaaaaa (approved):
👍

moritzzimmer (approved):

PR #61, opened 2020-10-05
feature / dependabot compatibility

vmk1vmk:
## What 

Refactor `Dockerfile` and `build_docker.sh` script to not contain any dependencies but rather outsource them to language specific package managers and build tools.

- `gradle` for Java
- `yarn` / `npm` for Node
- `go` for golang

## Why

We would like to get automatic version bump suggestions from Dependabot to remove the cognitive overhead of version checking for the developers of the `tapir` models.

## How

feature: added `build.gradle` file with task to download the linux executable from the java lib `io.grpc:protoc-gen-grpc-java`
refactor: move `Dockerfile` from project root to `docker` directory
feature: added `docker/package.json` file that specifies the required versions of the `grpc-tools` and `ts-protoc-gen` packages
feature: added `yarn.lock` file
feature: used gradle base image to download java dependencies in `Dockerfile`
feature: replaced explicit npm install commands with copying of the `package.json` and `yarn.lock` files and installing all dependencies from there
refactor: removed version specification from docker build script `build_docker.sh`
feature: added `go.mod` and `go.sum` files to pin go dependencies and their versions in an easily accessible location feature: tag docker image with GHCR path instead of plain name
feature: added way to install external go dependencies by using `go mod download` and then installing from the correct local path
refactor: better naming for docker helper containers
refactor: alphabetically ascending composition order
docs: documented instructions to bump versions and release the docker file
refactor: removed go grpc version from build script
feature: added `update` script to `package.json` to trivially update all dependencies
refactor: replaced yarn with npm by changing the `Dockerfile` instructions, removing the `yarn.lock` and adding the `package-lock.json` file
feature: added gradle version locking
docs: added step to generate lockfile when gradle dependencies where changed
refactor: merged docker and build specific dependency definitions for each language
refactor: removed `build_docker.sh` script for Dockerfile generation and added docker image creation and release to `Makefile`
refactor: removed outcommented code from `Dockerfile`
refactor: replaced `npm install` with `npm ci`


docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+RUN gradle download
+RUN mv installer /
+
+
+# Go
+FROM golang:1.15 as gopher
+ARG GO_GRPC_TOOLS_VERSION
+
+RUN mkdir /temp
+RUN cd /temp
```
harryurban, line 31: does nothing because each RUN is done in its own shell
just use WORKDIR

vmk1vmk: You start in the default `WORKDIR` which corresponds to hte `$GOPATH`, the default place where go dependencies are stored. Go throws an error when it sees an `go.mod` file in there, thats why this steps are required. @moritzzimmer  Do you have an opinion or a suggestion in that case?

docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+FROM node:lts-stretch as protoc-bin
+ARG PROTOC_VERSION
+RUN mkdir /installer
+RUN wget https://github.com/protocolbuffers/protobuf/releases/download/v$PROTOC_VERSION/protoc-$PROTOC_VERSION-linux-x86_64.zip -O protoc.zip
+RUN mkdir -p /installer/protoc
+RUN unzip -o protoc.zip -d /installer/protoc/
+
+
+# Node
+FROM node:lts-stretch as node
```
bobaaaaa, line 12:
```suggestion
FROM node:lts-alpine as node
```

29MB vs. 300MB

If this does not work try `node:lts-stretch-slim` (50MB)

docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+
+COPY package.json .
+COPY yarn.lock .
+RUN yarn install
```
bobaaaaa, line 16:
```suggestion
RUN npm install
```

docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+RUN wget https://github.com/protocolbuffers/protobuf/releases/download/v$PROTOC_VERSION/protoc-$PROTOC_VERSION-linux-x86_64.zip -O protoc.zip
+RUN mkdir -p /installer/protoc
+RUN unzip -o protoc.zip -d /installer/protoc/
+
+
+# Node
+FROM node:lts-stretch as node
+
+COPY package.json .
+COPY yarn.lock .
```
bobaaaaa, line 15:
```suggestion
COPY package-lock.json .
```

docker/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "grpc-tools": "1.9.1",
+    "ts-protoc-gen": "0.13.0"
+  }
+}
```
bobaaaaa, line 6:
```suggestion
{
  "dependencies": {
    "grpc-tools": "1.9.1",
    "ts-protoc-gen": "0.13.0"
  },
  "scripts": {
    "update": "npx npm-check --update"
  }
}
```

bobaaaaa (changes requested):
please use `npm` instead of `yarn`

bobaaaaa (approved):
node part looks good 👍

Makefile
```diff
@@ -61,5 +64,16 @@ clean: ## Deletes all generated files
 	rm -rf `find $(GO_DIR) -type d \( -iname "*" ! -iname "go.mod" ! -iname "go.sum" ! -iname "*_test.go" \) -mindepth 1 -maxdepth 1` 2> /dev/null || true
 	rm -rf `find $(NODE_DIR) -type d \( -iname "*" ! -iname "node_modules" ! -iname "__tests__" \) -mindepth 1 -maxdepth 1` 2> /dev/null || true
 
+image-build: ## Build new docker image
+	docker build \
+		-t ghcr.io/stroeer/protoc-dockerized:latest \
+		-t ghcr.io/stroeer/protoc-dockerized:$(PROTOC_VERSION) \
+		--build-arg PROTOC_VERSION=$(PROTOC_VERSION) .
```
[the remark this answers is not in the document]
harryurban: we noticed that locally on a devs macbook you wouldn't have a pull request or other incrementing numbers present that are set by github. The only way to reach that would be to modify files during a workflow run, which is not a good option in our opinion. 
We thought about creating a hash of the language-specific build files with a make target, but came to the conclusion that there is no need to keep old images of the same protoc version and would rollback incase of errors with smaller bumps and re-create the old image by releasing the image again.

harryurban (changes requested):
ship it

harryurban (approved):

PR #65, opened 2020-10-07
Author message specification

moritzzimmer:
This is a draft to discuss the proto message structure for authors which will be needed on article pages and on teasers (partly).

I discussed 2 options with @thisismana upfront:

- model authors as an `Element` with a new type `ElementType`
- model an author as a specific proto message type

I included code changes for the latter as a basis for discussion because:

- authors do have a numeric id in the cms which we might want to persist - elements don't have such a field
- authors do have a couple of fields like `first_name`, `last_name`, `vg_wort_id` etc. which could fit well into a future proof `map<string, string> fields` structure which we don't have on the element level. We could put it on the asset level though but let's discuss this

## persistence

As an outlook for the persistence: since we need at least a set of the author data on teasers and articles, it makes sense to fetch them via a single `dynamodb:GetItem`. As a consequence we might decide to denormalize the author data at transformation time into the `core.Article`. Changes in CMS  `data records` are communicated via stream. @Mi3-14159 and the CMS team still need to include affected articles in such an update message though.


thisismana:
`Authors` are not only used for the `Content`, but also for `Images`. There is field `source` within the media library which points to a author document.

Example:

A news article is imported and the author is set to `Author[id=dpa]`. Also, the images within the article `Article > Image > Media-Library` may have an author set.

How to handle this? `AuthorElement`/`AuthorAsset` would fit nicely for this use case?

harryurban:
maybe relevant, from requirements from Team Search:
```
Alle zusätzlichen Artikelbestandteile sollten aber zum Artikelexport als Listen zur Verfügung stehen, zum Beispiel integrierte Bilder oder Bildershows, Videos, Related Articles, integrierte oEmbeds, etc. 
```

harryurban:
```
{
  "elements": [{
    "type": ["ELEMENT_TYPE_AUTHOR", "ELEMENT_TYPE_IMAGE"],
    "relations": ["ELEMENT_RELATION_AUTHOR"],
    "assets": [{
      "type": "ASSET_TYPE_IMAGE",
      "fields": {
        "source": "https://dpa.de/authors/schwildy.png",
        "width": "400",
        "height": "400"
      }
    }, {
      "type": "ASSET_TYPE_AUTHOR",
      "fields": {
        "name": "Schwilden",
        "prename": "Freddy",
        "author_page": "/authors/schwildy",
        "vgwort": "/23434/foo.png"
      }
    }, {
      "type": "ASSET_TYPE_METADATA",
      "fields": {
        "caption": "Unser geiler Schwildt"
      }
    }]
  }, {
    "type": ["ELEMENT_TYPE_AUTHOR", "ELEMENT_TYPE_IMAGE"],
    "relations": ["ELEMENT_RELATION_AUTHOR"],
    "assets": [{
      "type": "ASSET_TYPE_IMAGE",
      "fields": {
        "source": "https://dpa.de/authors/omi.png",
        "width": "400",
        "height": "400"
      }
    }, {
      "type": "ASSET_TYPE_AUTHOR",
      "fields": {
        "name": "Michalsky",
        "prename": "Oliver",
        "author_page": "/authors/omig",
        "vgwort": "/5656/foo.png"
      }
    }, {
      "type": "ASSET_TYPE_METADATA",
      "fields": {
        "caption": "Stdv Redk. Oliver Michalsky"
      }
    }]
  }, {
    "type": ["ELEMENT_TYPE_AGENCY"],
    "relations": ["ELEMENT_RELATION_AGENCY"],
    "assets": [{
      "type": "ASSET_TYPE_AGENCY",
      "fields": {
        "name": "dpa"
      }
    }]
  }]
}
```

harryurban:
Think about usecases and limiations when using repeated ElementType

harryurban:
# status quo
Currently there are two main proposal, which  are similar but one could be migrated to the other later on:

## meaning of the value of `ElementType`
The `ElementType` defines:
- the primary expectable `AssetType` in an `Element`
  - e.g. `ImageElement` always has assets of type `ImageAsset`
- the expectable map entries in the fields map
  - e.g. `ImageElement` always has fields, like `altText`, that belong to an image

## proposal from MaNa:
### changes
- extend `enum ElementType` with value `ELEMENT_TYPE_AUTHOR = 6;`
- add Authors of Articles as `AuthorElement`s.
- add approriate Assets, which can also be ImageAssets for the profile fotos of editors

### pros
- stable and simple api
- doesn't change the original guardian model, of which the tapir model derived from

### cons
- consumers need to know about, which ElementTypes also "can be used as" an ImageElement (fulfil the meaning defined above)


## Harrys extension of MaNas proposal
### changes
- make ElementType and array
```
message Element {
  repeated ElementType type = 1;
...
}```

-> Elements that fulfilf the expectations regarding multiple ElementTypes will get the appropriate types

### pros
- removes the implicit knowledge about what ElementTypes should or can also be treated as an ImageElement
- keeps the logic about, which ElementTypes fulfill also other predicates, in one codebase
- makes filtering Elements by validity dates easier, by filtereing the Elements by the types, that can have Assets with validities

### cons
- breaks with the guardian model

### additional argumentation
proto3 doesn't allow polymorphy, but by having a field `type` in messages is proto3's way of expressing polymorphy, so that consumers can treat objects according to the type:
`is type X == is instanceof X`
making the ElementType a repeated field, just extends this polymorphism to
`is type X, Y == implements X,Y`

**Comment by bobaaaaa** [comment]
Thats a perfect summary 👍. I will throw a third proposal I like more:

### Recursive elements

```diff
message Element {
  ElementType type = 1;
  repeated ElementRelation relations = 2;
  repeated Asset assets = 3;
+ repeated Element children = 4;
}
```

#### pros

- non breaking
- fixed: Author element **has** an **optional** image element. _(instead of: Author is an image)_
- fixed: Video element **has** an **optional** poster image element. _(instead of: Video is a poster image)_
- fixed: Image gallery element **has** many image elements. _(instead of: Image gallery is many images)_
- a single `MetadataAsset` per element.
- no `index` needed to find the correct `MetadataAsset` for image galleries.

#### cons

- recursive structure
- breaks with the guardian model
- ???

bobaaaaa (approved):
👍

PR #67, opened 2020-10-07
CurationService V2

harryurban:
## add v2 curation service
- supports transporting labels from coremedia lists to display as stage headlines to the user
- adds batch service to get curation for all relevant lists in one call
  - can reduce risk for single failed calls
  - makes implementation in paperboy easier (no need to merge several `CompletableFuture`s)

## WHY
Make Version Zero for Homepage happen
- support dynamic purpose stages (aka `themenspezifische Bühnen`) with headlines defined by CvD / editors 

## Technical background about this change for polyphase service
- utilize property `toplineText` of Coremedia List objects for curated stage labels

## Discussion
move all coremedia throwaway code to polyphase by moving list logic to polyphase 
-> in paperboy request curation for `/` from polyphase
-> lists and their stage mappings would only be known in the polyphase service which should be deleted before golive


harryurban:
replaced by #83

PR #69, opened 2020-10-13
`uint64` must not be used, instead use `int64`

thisismana:


bobaaaaa (approved):

harryurban (approved):
🤷 if there are no unsingend integers in these languages I see that we maybe have to ensure positive values in code and switch to int64

the Authors id is also uint64 btw

thisismana:
> the Authors id is also uint64 btw

this is already fixed in the author pr #65

PR #80, opened 2020-10-14
Added new content sub types

thisismana:


bobaaaaa (approved):

harryurban:
Martin mentioned that there might be subtypes also for other maintypes besides text articles.
should we use own subtype Enums for each of the maintypes?
IIRC we only used one in the past project for subtypes across all maintypes, and in the end it felt like, that we could have deleted the maintype field and only use the subtype.

@thisismana

Issue #84, opened 2020-10-16
Add documentation generator


bobaaaaa:
At some point we need some documentation for some code conventions. Examples:

- list of valid `BodyNode.type` strings
- list of valid `fields` keys

### Requirements

- generate HTML documentation based on code comments
- deploy & host to GitHub pages (latest published version)

### Links

https://github.com/pseudomuto/protoc-gen-doc

### Future

- How do we deal with API changes? Versioned documentation?

moritzzimmer:
We (@thisismana, @bobaaaaa, @harryurban and @moritzzimmer) to first invest time in documenting our proto messages and gRPC APIs. Generating a HTML documentation in addition to the documented proto files might follow later.

PR #83, opened 2020-10-15
Curation Service for Homepage Version Zero

harryurban:
Refactor curation service for Homepage Version Zero in a breaking change
- remove required implicit knowledge about the cms behind the service
  - by speaking `stage_id`s between Api and CurationService.
  - Polyphase will map `stage_ids` to actual coremedia numeric list_ids
- add optional stage label from cm lists
- change implementation to a batch service
- strip down models to necessary data
- add documentating comments for autogenerated documentation

PR #87, opened 2020-10-23
tapir API docs wiki generator

harryurban:
# What
- [Dockerfile] add binary for protoc for generating docs from proto files
- [Makefile] add target to generate docs for all of /stroeer in one `api-spec.md`
- [CI] add workflow to add a link into the index file `api-spec.md` to the specs for the release

# Why
- Will produce this: https://github.com/stroeer/tapir/wiki/api-spec
- #84

# Comment
- generated docu relies on proper comments in files
  - see an example here:  https://github.com/stroeer/tapir/pull/83
- syntax for comments:
  - https://github.com/pseudomuto/protoc-gen-doc#writing-documentation

# Beware
- checks fail because docker image with the new binary was not pushed yet

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -1,4 +1,4 @@
-name: publish
+name: wiki-docs-generator
 
 on:
   release:
@@ -21,8 +21,8 @@ jobs:
           git config --local user.email "action@github.com"
           git config --local user.name "GitHub Action"
           cd tapir.wiki
-          cd tapir.wiki
-          [ -d api-specs/$RELEASE_TAG ] || mkdir -p api-specs/$RELEASE_TAG
+          [ -d api-spec
+          s/$RELEASE_TAG ] || mkdir -p api-specs/$RELEASE_TAG
           yes | cp ../api-docs/api-spec* ./api-specs/$RELEASE_TAG
           cd ./api-specs/$RELEASE_TAG
           echo "\n\n## $RELEASE_TAG \n" >> ../../api-spec.md
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -8,7 +8,8 @@ jobs:
   docs:
     env:
       RELEASE_TAG: ${{ github.event.release.tag_name }}
-      SPEC_FOLDER: api-specs/${{ github.event.release.tag_name }}
+      RELEASE_SPEC_FOLDER: api-docs/${{ github.event.release.tag_name }}
+      SPEC_FILE_NAME: api-spec.md
     steps:
       - uses: actions/checkout@v2
       - name: Gen docs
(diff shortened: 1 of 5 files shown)


.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,34 @@
...
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          cd tapir.wiki
+          [ -d $SPEC_FOLDER ] || mkdir -p $SPEC_FOLDER
+          yes | cp ../api-docs/api-spec* $SPEC_FOLDER
+          cd $SPEC_FOLDER
+          echo "\n\n## $RELEASE_TAG \n" >> ../../api-spec.md
+          for n in api-spec-stroeer*; do filepath=${n/.md/}; label=${filepath/api-spec-/}; echo "• [$label]($SPEC_FOLDER/$filepath)" >> ../../api-spec.md; done
+          cd ../..
```
harryurban, line 30: I am sure there is much easier and shorter code for this part. so feedback is very welcome.

Dockerfile
```diff
@@ -42,7 +48,8 @@ COPY --from=gopher /go/bin/protoc-gen-go-grpc /usr/bin/
 COPY --from=java /installer/protoc-gen-grpc-java /usr/bin/
 COPY --from=node /node_modules /node_modules
 COPY --from=protoc-bin /installer/protoc /usr/bin/protoc
-
+COPY --from=protoc-bin /installer/protoc-gen-doc /usr/bin/protoc-gen-doc
+RUN chmod +x /usr/bin/protoc-gen-grpc-java
 RUN chmod +x /usr/bin/protoc-gen-grpc-java
```
bobaaaaa, line 53: duplicate?

harryurban: Correct. Should be other filename but since it works like that I will remove it

Makefile
```diff
@@ -51,29 +53,40 @@ endif
...
 	[ -d $@ ] || mkdir -p $@
 
-stroeer/%: $(OUTPUT)
-	@echo "+ $@"
-	$(PROTOC) $(FLAGS) $(shell find $@ -iname "*.proto" -exec echo -n '"{}" ' \; | tr '\n' ' ')
+
+stroeer/%: $(OUTPUT) $(DOCS_DIR) clean-docs
+	@echo "+ compile: $@"
+	$(eval doc_name := api-spec-$(subst /,.,$@))
+	$(PROTOC) $(FLAGS) --doc_opt=markdown,$(doc_name).md $(shell find $@ -iname "*.proto" -exec echo -n '"{}" ' \; | tr '\n' ' ')
```
harryurban, line 67: Will change to an own target

bobaaaaa:
I already talked to Harry. I personally would prefer a github pages deployment (not into the wiki) because:

- its versioned as a whole
- its unchangeable

Makefile
harryurban: done

harryurban:
draft implementation pr is closed for now:
- `protoc-gen-doc` library is not up to date with proto3 (doesn't know `oneof` for example)
- usecases for a markdown docu are limited

maybe this will PR be replaced with a proper and robust solution as a hobby project later on

PR #107, opened 2020-11-12
grpc-gateway tooling

moritzzimmer:
Added tooling support for [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) and model for PoC of an article page service


Makefile
```diff
@@ -58,6 +58,14 @@ stroeer/%: $(OUTPUT)
 	@echo "+ $@"
 	$(PROTOC) $(FLAGS) $(shell find $@ -iname "*.proto" -exec echo -n '"{}" ' \; | tr '\n' ' ')
 
+gateway: ## Generates grpc-gateway resources
```
harryurban, line 61: so this is an example? or do we only want to generate the gateway resources for the article_service?

moritzzimmer: it's just a first step. We can and probably should also generate a gateway for the section service as well

PR #111, opened 2020-11-20
Documented core.Article proto messages

moritzzimmer:
@thisismana @bobaaaaa @harryurban @FalkoBasner in addition to the review of the docs we should also talk about the TODOs in the PR.

### DIFF
--- a/stroeer/core/v1/article.proto
+++ b/stroeer/core/v1/article.proto
@@ -39,14 +39,14 @@ message Article {
   // The entry set is defined by the content management system and
   // will vary depending on the main type of the article.
   //
-  // For CONTENT_TYPE_ARTICLE, this map will always contain non empty values
+  // For `CONTENT_TYPE_ARTICLE`, this map will always contain non empty values
   // for the following keys:
   //
   // * headline
   // * top_line
   // * summary
   //
-  // For CONTENT_TYPE_GALLERY, this map will always contain non empty values
+  // For `CONTENT_TYPE_GALLERY`, this map will always contain non empty values
   // for the following keys:
   //
   // * headline
@@ -57,7 +57,7 @@ message Article {
   // Recursive body structure of the article to render on detail
   // pages.
   //
-  // This message will be empty for the teaser representation of the article.
+  // This message will be empty for the teaser representation (e.g. section pages) of the article.
   Body body = 7;
 
   // Article metadata like publication state and technical timestamps.
@@ -65,7 +65,7 @@ message Article {
 
   // Elements of the article which are not part of the body, e.g. author information
   // opener image/video etc. Those elements should be used to render the article
-  // as a teaser e.g. on overview pages.
+  // as a teaser e.g. on section pages.
   repeated Element elements = 9;
 
   // Extracted entities from the article body like people, places, organizations etc.
@@ -115,7 +115,7 @@ message Metadata {
 // as a teaser e.g. on overview pages.
 //
 // Different types of elements like images or videos share
-// the same message structure distinguished by the ElementType field.
+// the same message structure distinguished by the `ElementType` field.
 message Element {
   // Type of the element.
   ElementType type = 1;
--- a/stroeer/core/v1/article.proto
+++ b/stroeer/core/v1/article.proto
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
 option java_multiple_files = true;
 option java_package = "de.stroeer.core.v1";
 
+// An article represents a piece of content created in the content management
+// system. Different types of content like text or video articles share
+// the same message structure distinguished by the ContentType field.
+//
+// Depending on the service used to retrieve an article, this message might
+// only contain teaser relevant information (e.g. with no Body field).
+//
+// Message fields and map entries are optional unless commented otherwise.
 message Article {
+  // Unique ID of the article defined by the content management system (required).
   int64 id = 1;
+
+  // Main content type of the article.
   ContentType type = 2;
+
+  // Subtype of the article.
   ContentSubType sub_type = 3;
+
+  // References to get access to the article on different technical platforms like
+  // canonical web or AMP (required).
   References references = 4;
+
+  // Hierarchical section information of the article (required).
   Section section = 5;
+
+  // Generic map containing general content and configuration information of
+  // the article (required).
+  //
+  // The entry set is defined by the content management system and
+  // will vary depending on the main type of the article.
+  //
+  // For CONTENT_TYPE_ARTICLE, this map will always contain non empty values
+  // for the following keys:
+  //
+  // * headline
+  // * top_line
+  // * summary
+  // * meta_title
+  //
+  // For CONTENT_TYPE_GALLERY, this map will always contain non empty values
+  // for the following keys:
+  //
+  // * headline
+  //
+  // Clients must be resilient to unknown or missing entry sets.
   map<string, string> fields = 6;
+
+  // Recursive body structure of the article to render on detail
+  // pages (optional).
+  //
+  // This message will be empty for the teaser representation of the article.
+  //
+  // see https://github.com/orgs/stroeer/teams/api/discussions/1
   Body body = 7;
+
+  // Technical metadata of the article (optional).
   Metadata metadata = 8;
+
+  // Elements of the article to render as a teaser e.g. on overview pages (optional).
   repeated Element elements = 9;
+
+  // Entities of the article like persons, places etc. (optional).
   repeated string keywords = 10;
+
+  // Unique IDs of articles related to this article (optional).
+  //
+  // Related articles are defined manually in the content management system
+  // by the editorial department.
   repeated int64 onwards = 12;
 }
 
+// Technical metadata of the article.
 message Metadata {
+
+  // State of the article in the content management system (optional).
+  // see https://github.com/stroeer/tapir/issues/64
   string state = 1;
-  //  editorial date: Gültig von
+
+  // Manually set editorial date ('Gültig von') at which the article is valid to deliver on
+  // digital platforms in seconds of UTC time since Unix epoch (optional).
   google.protobuf.Timestamp valid_from = 2;
-  //  editorial date: Gültig bis
+
+  // Manually set editorial date ('Gültig bis') till the article is valid to deliver on
+  // digital platforms in seconds of UTC time since Unix epoch (optional).
   google.protobuf.Timestamp valid_to = 3;
-  //  editorial date: Zuletzt bearbeitet
+
+  // editorial date: Zuletzt bearbeitet
+  // TODO: this seems unused....remove?
   google.protobuf.Timestamp last_edited = 4;
+
   //  editorial date: Zuletzt publiziert
+  // TODO: this seems unused....remove?
   google.protobuf.Timestamp last_published = 5;
+
   // editorial date: Erstellungsdatum
+  // TODO: this seems unused....remove?
   google.protobuf.Timestamp created = 6;
-  // editorial date: Aktualisierungsdatum
+
+  // Manually set editorial date ('Aktualisierungsdatum') at which the article
+  // was updated in seconds of UTC time since Unix epoch (optional).
   google.protobuf.Timestamp updated = 7;
 }
 
+// Elements of the article to render as a teaser e.g. on overview pages.
+//
+// Different types of elements like images or videos share
+// the same message structure distinguished by the ElementType field.
 message Element {
+  // Type of the element.
   ElementType type = 1;
+
+  // TODO: this might be obsolete with https://github.com/orgs/stroeer/teams/api/discussions/1
   repeated ElementRelation relations = 2;
+
+  // Assets of the element like images or videos (optional).
   repeated Asset assets = 3;
+
+  // Children of the element (optional).
   repeated Element children = 4;
 }
 
+// Asset of an element.
 message Asset {
+  // Type of the asset.
   AssetType type = 1;
+
+  // TODO: do we still need the index?
   int32 index = 2;
+
+  // Generic map containing general content and configuration information of
+  // the asset (optional).
+  //
+  // Depending on the type of the asset, this map will contain different
+  // entry sets, e.g. 'source', 'width' and 'height' for images.
+  //
+  // Clients must be resilient to unknown or missing entry sets.
   map<string, string> fields = 3;
+
+  // Technical metadata of the asset (optional).
   Metadata metadata = 4;
 }
 
+// Hierarchical section information of the article till root.
+//
+// For example:  /politik/deutschland/ -> /politik/ -> /
 message Section {
+  // Section path of the article, e.g. /politik/deutschland/ (required).
   string path = 1;
+
+  // Label of the section path, e.g. "Politik"  (TODO: required or optional?).
   string label = 2;
+
+  // Parent section (optional).
   Section parent = 3;
 }
 
+// References to get access to the article on different technical platforms like
+// canonical web or AMP.
 message References {
-  // article's path, e.g. /section/id_$ID/title.html
+  // Path of the article's path, e.g. /section/id_$ID/title.html (required).
   string path = 1;
-  // article's canonical URL, may differ if external, e.g. https://www.giga.de/external.html
+
+  // Canonical URL of the article, may differ if external, e.g. https://www.giga.de/external.html
+  // (required).
   string canonical = 2;
-  // article's AMP URL, no pattern specified yet.
+
+  // AMP URL of the article (optional).
   string amp = 3;
 }
 
+// Body of the article to render on detail pages.
+//
+// see https://github.com/orgs/stroeer/teams/api/discussions/1
 message Body {
+  // Recursive Body nodes.
   repeated BodyNode children = 1;
 }
 
+// Recursive body structure representing all types of possible
+// body nodes inside an article.
+//
+// Clients must be resilient to unknown or missing nodes.
 message BodyNode {
+  // Type of the node, e.g. 'a', 'p', 'ul', 'li', 'text' (required).
   string type = 1;
+
+  // Text of the node (optional).
   string text = 2;
+
+  // Additional information for the node depending on it's type,
+  // e.g. 'href' for 'a' nodes (optional).
   map<string, string> fields = 3;
+
+  // Children of the node, e.g. the 'text' of a 'p' (optional).
   repeated BodyNode children = 4;
-  repeated Element elements = 5; // inline-video, content|rich-links, gallery
+
+  // Elements of the node, e.g. inline videos or galleries (optional).
+  repeated Element elements = 5;
 }
 
+// Main content type of the article.
 enum ContentType {
+  // Not specified.
   CONTENT_TYPE_UNSPECIFIED = 0;
+
+  // A text article.
   CONTENT_TYPE_ARTICLE = 1;
+
+  // TODO: do we have/need images as main types?
   CONTENT_TYPE_IMAGE = 2;
+
+  // A video article.
   CONTENT_TYPE_VIDEO = 3;
+
+  // A gallery article.
   CONTENT_TYPE_GALLERY = 4;
+
+  // An oembed article.
   CONTENT_TYPE_OEMBED = 5;
+
+  // TODO: do we have/need authors as main types?
   CONTENT_TYPE_AUTHOR = 6;
+
+  // TODO: do we have/need agencies as main types?
   CONTENT_TYPE_AGENCY = 7;
 }
 
+// Subtype of the article.
 enum ContentSubType {
+  // Not specified.
   CONTENT_SUB_TYPE_UNSPECIFIED = 0;
+
   // Meldung/Nachricht — this is the default
   CONTENT_SUB_TYPE_NEWS = 1;
-  // Kolumne
+
+  // 'Kolumne'
   CONTENT_SUB_TYPE_COLUMN = 2;
-  // Bericht
+
+  // 'Bericht'
   CONTENT_SUB_TYPE_REPORT = 3;
-  // Kommentar
+
+  // 'Kommentar'
   CONTENT_SUB_TYPE_COMMENTARY = 4;
-  // Interview
+
+  // 'Interview'
   CONTENT_SUB_TYPE_INTERVIEW = 5;
-  // Pro und Kontra/Streitgespräch
+
+  // 'Pro und Kontra/Streitgespräch'
   CONTENT_SUB_TYPE_CONTROVERSY = 6;
-  // Reportage
+
+  // 'Reportage'
   CONTENT_SUB_TYPE_STORY = 7;
-  // Rezension/Kritik
+
+  // 'Rezension/Kritik'
   CONTENT_SUB_TYPE_RECENSION = 8;
-  // Tagesanbruch
+
+  // 'Tagesanbruch'
   CONTENT_SUB_TYPE_TAGESANBRUCH = 9;
-  // Analyse
+
+  // 'Analyse'
   CONTENT_SUB_TYPE_ANALYSIS = 10;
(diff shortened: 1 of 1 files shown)


moritzzimmer:
### Remarks from the google api linter

- https://linter.aip.dev/203/optional

_This rule looks at any field with “optional” (or similar forms) in the comment, and complains if it does not have a google.api.field_behavior annotation._

> Note: It is generally recommended not to document “optional” at all

Shall we just remove all `(optional)` statements?

moritzzimmer:
- https://google.aip.dev/142

_Timestamp fields should end in `_time`._

Ignore? Rename?

FalkoBasner:
> ### Remarks from the google api linter
> * https://linter.aip.dev/203/optional
> 
> _This rule looks at any field with “optional” (or similar forms) in the comment, and complains if it does not have a google.api.field_behavior annotation._
> 
> > Note: It is generally recommended not to document “optional” at all
> 
> Shall we just remove all `(optional)` statements?

So in proto3 everything is more or less optional. 
https://stackoverflow.com/questions/31801257/why-required-and-optional-is-removed-in-protocol-buffers-3
Yeah I would just remove optional in comments. In a way every API user can decide on its own what information is need or not.

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
   CONTENT_SUB_TYPE_ANALYSIS = 10;
-  // Satire/Glosse
+
+  // 'Satire/Glosse'
   CONTENT_SUB_TYPE_SATIRE = 11;
 }
 
+// Type of an element.
 enum ElementType {
+  // Not specified.
```
thisismana, line 279:
```suggestion
  // Not specified, not supported or yet unknown.
```

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  // A gallery article.
   CONTENT_TYPE_GALLERY = 4;
+
+  // An oembed article.
   CONTENT_TYPE_OEMBED = 5;
+
+  // TODO: do we have/need authors as main types?
   CONTENT_TYPE_AUTHOR = 6;
+
+  // TODO: do we have/need agencies as main types?
```
thisismana, line 234: this is more due to technical reasons, when we fetch the author/agency data record, this is used to parse and map the type within the adapter. This type is never used for actual content:

```javascript
{
    "systemdata": {
        "contentType": "agency",
        "documentType": "data-record"
    }
}
```

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
 
+// Recursive body structure representing all types of possible
+// body nodes inside an article.
+//
+// Clients must be resilient to unknown or missing nodes.
 message BodyNode {
+  // Type of the node, e.g. 'a', 'p', 'ul', 'li', 'text' (required).
   string type = 1;
+
+  // Text of the node (optional).
```
thisismana, line 197:
```suggestion
  // Text of the node (optional), only set for text nodes (`type == 'text'`).
```

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+//
+// For example:  /politik/deutschland/ -> /politik/ -> /
 message Section {
+  // Section path of the article, e.g. /politik/deutschland/ (required).
   string path = 1;
+
+  // Label of the section path, e.g. "Politik"  (TODO: required or optional?).
   string label = 2;
+
+  // Parent section (optional).
```
thisismana, line 163:
```suggestion
  // Parent section (optional). Usually points to the parent section, if there is none, e.g. when in `root` this value is `null`
```

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+// Different types of elements like images or videos share
+// the same message structure distinguished by the ElementType field.
 message Element {
+  // Type of the element.
   ElementType type = 1;
+
+  // TODO: this might be obsolete with https://github.com/orgs/stroeer/teams/api/discussions/1
   repeated ElementRelation relations = 2;
+
+  // Assets of the element like images or videos (optional).
```
thisismana, line 125: I think this is not optional, every element has at least one asset (metadata) that describes the element.

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
 
+// Technical metadata of the article.
 message Metadata {
+
+  // State of the article in the content management system (optional).
+  // see https://github.com/stroeer/tapir/issues/64
   string state = 1;
-  //  editorial date: Gültig von
+
+  // Manually set editorial date ('Gültig von') at which the article is valid to deliver on
```
thisismana, line 89: For the technical details of timestamps (such as UTC, Unixtime etc) I'd refer the reader to the [docs](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/timestamp)

> A Timestamp represents a point in time independent of any time zone or calendar, represented as seconds and fractions of seconds at nanosecond resolution in UTC Epoch time.

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
   google.protobuf.Timestamp valid_from = 2;
-  //  editorial date: Gültig bis
+
+  // Manually set editorial date ('Gültig bis') till the article is valid to deliver on
+  // digital platforms in seconds of UTC time since Unix epoch (optional).
   google.protobuf.Timestamp valid_to = 3;
-  //  editorial date: Zuletzt bearbeitet
+
+  // editorial date: Zuletzt bearbeitet
+  // TODO: this seems unused....remove?
```
thisismana, line 98: many of the date fields vanished from the CMS lately, this should be dealt with in a separate PR.

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  // The entry set is defined by the content management system and
+  // will vary depending on the main type of the article.
+  //
+  // For CONTENT_TYPE_ARTICLE, this map will always contain non empty values
+  // for the following keys:
+  //
+  // * headline
+  // * top_line
+  // * summary
+  // * meta_title
```
thisismana, line 48:
```suggestion
```

thisismana: this is not mandatory

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+//
+// Message fields and map entries are optional unless commented otherwise.
 message Article {
+  // Unique ID of the article defined by the content management system (required).
   int64 id = 1;
+
+  // Main content type of the article.
   ContentType type = 2;
+
+  // Subtype of the article.
```
thisismana, line 26:
```suggestion
  // Subtype of the article (optional). For `CONTENT_TYPE_ARTICLE` this field holds a sub type, for others like `CONTENT_TYPE_GALLERY` it may not.
```

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
 option java_multiple_files = true;
 option java_package = "de.stroeer.core.v1";
 
+// An article represents a piece of content created in the content management
+// system. Different types of content like text or video articles share
+// the same message structure distinguished by the ContentType field.
```
thisismana, line 13:
```suggestion
// the same message structure distinguished by the `Article.type` field.
```

FalkoBasner:
> ### Remarks from the google api linter
> * https://linter.aip.dev/203/optional
> 
> _This rule looks at any field with “optional” (or similar forms) in the comment, and complains if it does not have a google.api.field_behavior annotation._
> 
> > Note: It is generally recommended not to document “optional” at all
> 
> Shall we just remove all `(optional)` statements?

So in proto3 everything is more or less optional. 

> ### Remarks from the google api linter
> * https://google.aip.dev/142, _Timestamp fields should end in `_time`._
> * https://linter.aip.dev/142/time-field-names, _Use the imperative mood and a `_time` suffix for timestamps._
> 
> Ignore? Rename?

Well :-(, if the Linter says so. Maybe that helps the API in the long run. RENAME

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+//
+// Depending on the service used to retrieve an article, this message might
+// only contain teaser relevant information (e.g. with no Body field).
```
thisismana, line 16:
```suggestion
// only contain data required on overview pages (e.g. with `Article.body` set to `null` thus not containing any data that is only required on detail pages). This is sometimes referred to as _Teaser_.
```

FalkoBasner (approved):

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,318 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  // TODO: do we still need the index?
   int32 index = 2;
+
+  // Generic map containing general content and configuration information of
+  // the asset (optional).
+  //
+  // Depending on the type of the asset, this map will contain different
+  // entry sets, e.g. 'source', 'width' and 'height' for images.
+  //
+  // Clients must be resilient to unknown or missing entry sets.
```
thisismana, line 146: we should create a gallery of samples for the most important elements as a reference.

thisismana (changes requested):

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+
+  // Parent section (optional). Usually points to the parent section, if there is none, e.g. when in `root` this value is `null`
   Section parent = 3;
 }
 
+// References to get access to the article on different technical platforms like
+// canonical web or AMP.
 message References {
-  // article's path, e.g. /section/id_$ID/title.html
+  // Path of the article's path, e.g. /section/id_$ID/title.html (required).
```
thisismana, line 169:
```suggestion
  // URL path for this article e.g. /section/id_$ID/title.html (required).
```

stroeer/core/v1/article.proto
moritzzimmer: see #112

stroeer/core/v1/article.proto
moritzzimmer: we should we put those examples? directly in the comment?

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  //
+  // Clients must be resilient to unknown or missing entry sets.
   map<string, string> fields = 6;
+
+  // Recursive body structure of the article to render on detail
+  // pages.
+  //
+  // This message will be empty for the teaser representation of the article.
+  //
+  // see https://github.com/orgs/stroeer/teams/api/discussions/1
```
harryurban, line 62: Discussion is in a gh team only accessible to members of Team Buzz. 
I understand the motivation behind this comment, but would still remove it before merging.

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
   Body body = 7;
+
+  // Technical metadata of the article.
```
harryurban, line 65: Comment does not add any useful information in my opinion.

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
   Metadata metadata = 8;
+
+  // Elements of the article to render as a teaser e.g. on overview pages.
```
harryurban, line 68: Elements are more than only things needed for a teaser, no?
Authors for example.

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
   repeated Element elements = 9;
+
+  // Entities of the article like persons, places etc.
```
harryurban, line 71:
```suggestion
  // Extracted entities like people, places, organizations, etc. 
```

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
   repeated string keywords = 10;
+
+  // Unique IDs of articles related to this article.
```
harryurban, line 74: Uniqueness is not guaranteed, I think?

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  //
+  // Related articles are defined manually in the content management system
+  // by the editorial department.
   repeated int64 onwards = 12;
 }
 
+// Technical metadata of the article.
```
harryurban, line 81:
```suggestion
// Article metadata such as publication state and date dependent validities.
```

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
 message Metadata {
+
+  // State of the article in the content management system.
+  // see https://github.com/stroeer/tapir/issues/64
   string state = 1;
-  //  editorial date: Gültig von
+
+  // Manually set editorial date ('Gültig von') at which the article is valid to deliver on
```
harryurban, line 88:
```suggestion
  // Manually set editorial date ('Gültig von'). Defines the start of a time based validity; it can derive from 3rd party licenses and can also be used to plan the release of an article without involvement of an editor. 
```

stroeer/core/v1/article.proto
harryurban: Or generate html / markdown docs. 😉

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  // Generic map containing general content and configuration information of
+  // the asset.
+  //
+  // Depending on the type of the asset, this map will contain different
+  // entry sets, e.g. 'source', 'width' and 'height' for images.
+  //
+  // Clients must be resilient to unknown or missing entry sets.
   map<string, string> fields = 3;
+
+  // Technical metadata of the asset.
```
harryurban, line 148: Validity data ?

harryurban (changes requested):

stroeer/core/v1/article.proto
shobhitsharma: All dates that are stored by us are CET or UTC?

moritzzimmer: UTC @shobhitsharma

stroeer/core/v1/article.proto
moritzzimmer: @harryurban then we should use a similar comments for `valid_to`

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  // Hierarchical section information of the article (required).
   Section section = 5;
+
+  // Generic map containing general content and configuration information of
+  // the article (required).
+  //
+  // The entry set is defined by the content management system and
+  // will vary depending on the main type of the article.
+  //
+  // For CONTENT_TYPE_ARTICLE, this map will always contain non empty values
```
bobaaaaa, line 42:
```suggestion
  // For `CONTENT_TYPE_ARTICLE`, this map will always contain non empty values
```

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  // for the following keys:
+  //
+  // * headline
+  // * top_line
+  // * summary
+  //
+  // For CONTENT_TYPE_GALLERY, this map will always contain non empty values
```
bobaaaaa, line 49:
```suggestion
  // For `CONTENT_TYPE_GALLERY`, this map will always contain non empty values
```

stroeer/core/v1/article.proto
moritzzimmer: But do we need those in the model then @thisismana ?

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  // canonical web or AMP (required).
   References references = 4;
+
+  // Hierarchical section information of the article (required).
   Section section = 5;
+
+  // Generic map containing general content and configuration information of
+  // the article (required).
+  //
+  // The entry set is defined by the content management system and
```
bobaaaaa, line 39: Is it worth to mention that flags are prefixed with `flag:`? cc @thisismana 

```json5
"flag:external": true
```

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  //
+  // * headline
+  //
+  // Clients must be resilient to unknown or missing entry sets.
   map<string, string> fields = 6;
+
+  // Recursive body structure of the article to render on detail
+  // pages.
+  //
+  // This message will be empty for the teaser representation of the article.
```
bobaaaaa, line 60:
```suggestion
  // This message will be empty for the teaser representation (e.g. section pages) of the article.
```

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
   Body body = 7;
+
+  // Article metadata like publication state and technical timestamps.
```
bobaaaaa, line 63: nullable?

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
   Metadata metadata = 8;
+
+  // Elements of the article which are not part of the body, e.g. author information
+  // opener image/video etc. Those elements should be used to render the article
+  // as a teaser e.g. on overview pages.
```
bobaaaaa, line 68:
```suggestion
  // as a teaser e.g. on section pages.
```

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  // was updated in seconds of UTC time since Unix epoch.
   google.protobuf.Timestamp updated = 7;
 }
 
+// Elements of the article which are not part of the body, e.g. author information
+// opener image/video etc. Those elements should be used to render the article
+// as a teaser e.g. on overview pages.
+//
+// Different types of elements like images or videos share
+// the same message structure distinguished by the ElementType field.
```
bobaaaaa, line 118:
```suggestion
// the same message structure distinguished by the `ElementType` field.
```

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  // Type of the element.
   ElementType type = 1;
+
+  // TODO: this might be obsolete with https://github.com/orgs/stroeer/teams/api/discussions/1
   repeated ElementRelation relations = 2;
+
+  // Assets of the element like images, videos or metadata (required).
   repeated Asset assets = 3;
+
+  // Children of the element.
```
bobaaaaa, line 129: Adding an example would help here:

- A author element might have an image element child
- A gallery element have image element children

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  // Recursive Body nodes.
   repeated BodyNode children = 1;
 }
 
+// Recursive body structure representing all types of possible
+// body nodes inside an article.
+//
+// Clients must be resilient to unknown or missing nodes.
 message BodyNode {
+  // Type of the node, e.g. 'a', 'p', 'ul', 'li', 'text' (required).
```
bobaaaaa, line 193: We need all of them here. Current paper impl.:

```ts
export type ApiBodyNodeType =
  | 'text'
  | 'p'
  | 'sub_headline'
  | 'a'
  | 'strong'
  | 'em'
  | 'ul'
  | 'ol'
  | 'li'
  | 'img'
  | 'article_sources'
  | 'article_source';
```

bobaaaaa: The last two are "special". We need to explain them.

stroeer/core/v1/article.proto
```diff
@@ -8,133 +8,317 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
   string type = 1;
+
+  // Text of the node, only set for text nodes (`type == 'text'`).
```
bobaaaaa, line 196: Just for clarification: the grpc null  value (== empty string) is still a valid body node? Do not filter body (type text) nodes with empty strings? Does this makes sense? cc @thisismana

bobaaaaa (commented):
Always "section" and not overview pages. And I would prefer to use always \`foo\` instead of 'foo' style

stroeer/core/v1/article.proto
thisismana: flag as `internal`?

stroeer/core/v1/article.proto
thisismana: As for gRPC, `""` is the default value, there is no `null`.
When transforming stuff from the CMS, I have options to strip out the whole `Node` if there is a leaf-node with blank text.

stroeer/core/v1/article.proto
thisismana: I'd also thought about a way to document field contents, either through fixtures or in the docs.
But all this is as volatile as is the CMS model.

stroeer/core/v1/article.proto
thisismana: `for i in 1,2,3 ; do grpcurl $API > docs/$i.buff; done`

stroeer/core/v1/article.proto
moritzzimmer: @bobaaaaa shall we remove the last 2 for now, since they might end up somewhere else in the model anyway?

stroeer/core/v1/article.proto
moritzzimmer: https://github.com/stroeer/tapir/issues/114

stroeer/core/v1/article.proto
moritzzimmer: if not documented as `required` all fields are optional (aka. have their default values)

bobaaaaa (approved):
This is a good start. lets iterate 👍

PR #150, opened 2021-02-04
feat(model): shared page data

### LINKED ISSUE #149: page: SectionPage with References for canonical link
## What

Currently only the `Article` has a `References` with the canonical link. For SEO reasons we also need this on all section pages.

## Proposal

```diff
message SectionPage {
   repeated Stage stages = 1;
+  References references = 2;
}

message ArticlePage {
   stroeer.core.v1.Article article = 1;
   repeated RelatedArticle related_articles = 2;
+  References references = 3; // because of same logic for `Pages` in FE
}
```

The only downside of this is that the `Article` and `ArticlePage` has the same `References` (duplication).

bobaaaaa:
Introducing a new message type for `SectionPage` to access all kind data to render a page's meta data or shared content. This message type contains the same data as the `Article` message. For this reason I renamed the message type `Section` to `SectionTree` to align the page field wordings.

```proto
// only for clarification:

message Article {
  References references = ?;
  SectionTree section_tree = ?;
  // ... more 
}

message Section {
  References references = ?;
  SectionTree section_tree = ?;
  // ... more 
}

message ArticlePage {
  Article article = 1;
  // ... more 
}

message SectionPage {
  Section section = 1;
  // ... more 
}
```

fixes #149

### DIFF
--- a/stroeer/core/v1/shared.proto
+++ b/stroeer/core/v1/shared.proto
@@ -1,0 +1,36 @@
+syntax = "proto3";
+
+package stroeer.core.v1;
+
+option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
+option java_multiple_files = true;
+option java_package = "de.stroeer.core.v1";
+
+
+// References to get access to the article on different technical platforms like
+// canonical web or AMP.
+message References {
+  // URL path for this article e.g. /section/id_$ID/title.html (required).
+  string path = 1;
+
+  // Canonical URL of the article, may differ if external, e.g. https://www.giga.de/external.html
+  // (required).
+  string canonical = 2;
+
+  // AMP URL of the article.
+  string amp = 3;
+}
+
+// Hierarchical section information of the article till root.
+//
+// For example:  /politik/deutschland/ -> /politik/ -> /
+message SectionTree {
+  // Section path of the article, e.g. /politik/deutschland/ (required).
+  string path = 1;
+
+  // Label of the section path, e.g. "Politik".
+  string label = 2;
+
+  // Parent section. Usually points to the parent section, if there is none, e.g. when in `root` this value is `null`
+  SectionTree parent = 3;
+}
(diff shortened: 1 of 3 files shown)


stroeer/core/v1/shared.proto
```diff
@@ -0,0 +1,36 @@
...
+  string canonical = 2;
+
+  // AMP URL of the article.
+  string amp = 3;
+}
+
+// Hierarchical section information of the article till root.
+//
+// For example:  /politik/deutschland/ -> /politik/ -> /
+message SectionTree {
```
bobaaaaa, line 27: Some thoughts:
While I think about the internal link structure I thought about the main navigation and how it could be handled by our models. The `SectionTree` would be perfect for this when the name would be more generic.

moritzzimmer (approved):

stroeer/core/v1/shared.proto
moritzzimmer: we won't need the hierarchy in the navigation though, do we? Still we could reuse this message - any naming ideas in mind @bobaaaaa ?

bobaaaaa: The most obvious name would be:

```proto
message ReferenceTree {
  // Section path of the article, e.g. /politik/deutschland/ (required).
  Reference reference = 1;

  // Parent section. Usually points to the parent section, if there is none, e.g. when in `root` this value is `null`
  ReferenceTree parent = 2;
}
```

If we take it 100% correctly even `Tree` is the wrong postfix -- it's an `TreeItem` or `Leaf`. /cc @thisismana

bobaaaaa: `LinkedReference` 🎉

thisismana (approved):

thisismana:
This will probably never go live, but I'm merging this as an intermediate towards upcoming changes proposed here: https://github.com/stroeer/tapir/blob/feat--Reference/Reference.proto

PR #155, opened 2021-02-10
feat: Reference

### LINKED ISSUE #121: stages: references
We need a concept for links (`<a href />`) as part of a stage ("Mehr zu Sport").

bobaaaaa:
This is it 🎉 

```proto
message Reference {
  string type = 1;
  string label = 2;
  string href = 3;
  map<string, string> fields = 4;
  repeated Reference children = 5;
}
```

### Features

- Usable as a single ref or multiple ref structures (like navigation) 
- Handling multiple `Reference`'s via `type` (like `Element`)
- Replaces `SectionTree` message

### Refactorings

- (💥) Deleted `SectionTree` message and used `Reference` instead + switched from bottom-top to top-bottom structure.
- (💥) Deleted `References` message (misleading naming). Now part of `fields`
- (💥) Reassigned article proto keys
- (✅) `Reference` in `Stage`'s
- (✅) API docs update formatting

### Why

Fixes #121


stroeer/core/v1/shared.proto
```diff
@@ -6,31 +6,41 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+// an anchor element.
+message Reference {
+  // The type is only needed for filtering multiple links of the same type in
+  // list of references.
+  //
+  // Example entries:
+  // * `title`
+  // * `navigation`
+  // * `sub_navigation`
+  string type = 1;
```
bobaaaaa, line 19: FYI: This `type` is not an enum because of total diverse usage. It depends on the context or integration type.

stroeer/page/section/v1/section_page.proto
```diff
@@ -17,12 +17,20 @@ message SectionPage {
...
-  // canonical web or AMP (required).
-  stroeer.core.v1.References references = 1;
+  // Generic map containing general content and configuration information of
+  // the section (required).
+  //
+  // This map will always contain non empty values for the following keys:
+  //
+  // * `url_path`: URL path for this section e.g. /sport/fussball/bundesliga/
+  // * `url_canonical`: Canonical URL of the section, may differ if external, e.g. https://www.giga.de/tech/
+  // * `url_amp`: AMP URL of the section
```
bobaaaaa, line 27: We could talk about this mandatory field names

stroeer/core/v1/shared.proto
```diff
@@ -6,31 +6,41 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+// For example:
+// `/` -> `/sport/` -> `/sport/fussball/`
+// [?] The name `Category` is used here to align namings with the CMS.
+message Category {
+  // Reference of the current category e.g. /politik/deutschland/ (required).
+  Reference reference = 1;
+
+  // All children of this category.
+  repeated Category children = 2;
 }
```
bobaaaaa, line 46: `top-to-bottom` vs. `bottom-to-top`

stroeer/core/v1/article.proto
```diff
@@ -29,12 +29,8 @@ message Article {
   // Subtype of the article. For `CONTENT_TYPE_ARTICLE` this field holds a sub type, for others like `CONTENT_TYPE_GALLERY` it may not.
   ContentSubType sub_type = 3;
 
-  // References to get access to the article on different technical platforms like
-  // canonical web or AMP (required).
-  stroeer.core.v1.References references = 4;
-
   // Hierarchical section tree information of the article (required).
-  stroeer.core.v1.SectionTree section_tree = 5;
+  stroeer.core.v1.Category section_tree = 5;
```
bobaaaaa, line 33: I still like `section_tree` as field name.

stroeer/core/v1/shared.proto
thisismana: we've discussed: if we do the sub navigation `top-to-bottom`, we should do the same for the `section_tree` (breadcrumb). So always start at the `root` and have a chain of `Category` with exactly one child each until the `bottom` is reached.

stroeer/core/v1/article.proto
thisismana: this does not matter on the (protobuf) wire, anyway.

thisismana (approved):

stroeer/core/v1/shared.proto
```diff
@@ -6,31 +6,41 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
-  string path = 1;
-
-  // Canonical URL of the article, may differ if external, e.g. https://www.giga.de/external.html
-  // (required).
-  string canonical = 2;
-
-  // AMP URL of the article.
-  string amp = 3;
+// Represents a single Reference or Link incl. all needed attributes to built
+// an anchor element.
```
moritzzimmer, line 10:
```suggestion
// A Reference represents a link to another entity - for example an `Article`, a `SectionPage` or an external website. 
```

stroeer/core/v1/shared.proto
moritzzimmer: @bobaaaaa can you given an example where we would use the type for filtering?

stroeer/core/v1/shared.proto
```diff
@@ -6,31 +6,41 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  // The type is only needed for filtering multiple links of the same type in
+  // list of references.
+  //
+  // Example entries:
+  // * `title`
+  // * `navigation`
+  // * `sub_navigation`
+  string type = 1;
+
+  // The text or label of the link.
```
moritzzimmer, line 21:
```suggestion
  // The text or label of the reference.
```

stroeer/core/v1/shared.proto
```diff
@@ -6,31 +6,41 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
+  string text = 2;
+
+  // Contains all optional attributes of a link.
```
moritzzimmer, line 24:
```suggestion
  // Contains all optional attributes of the reference.
```

stroeer/core/v1/shared.proto
```diff
@@ -6,31 +6,41 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
...
-
-  // Label of the section path, e.g. "Politik".
-  string label = 2;
-
-  // Parent section. Usually points to the parent section, if there is none, e.g. when in `root` this value is `null`
-  SectionTree parent = 3;
+// For example:
+// `/` -> `/sport/` -> `/sport/fussball/`
+// [?] The name `Category` is used here to align namings with the CMS.
+message Category {
```
moritzzimmer, line 40: IMO we shouldn't refer to the CMS for naming but decide which name is most suitable for our API and it's clients. We currently mix `Category` and `section_tree` in message name, comments and attribute names which can be confusing. 

I kind of liked the `tree` inside the naming since it represents a tree like structure.

stroeer/page/section/v1/section_page.proto
moritzzimmer: Maybe I need to see an example, but my first thought was that those URLs are represented by the `Category` with it's reference?!

stroeer/core/v1/shared.proto
bobaaaaa: Multiple links in a `Stage` or a custom page header.

```ts
// Instead of:
config.subNavigation
config.sponsoringReference
config.titleReference
// We want:
config.references.filter(isSubNavigation)
config.references.filter(isTitle)
config.references.filter(isSponsoring)
```

stroeer/page/section/v1/section_page.proto
bobaaaaa: I removed the message `References`:

```proto
// References to get access to the article on different technical platforms like
// canonical web or AMP.
message References {
  // URL path for this article e.g. /section/id_$ID/title.html (required).
  string path = 1;

  // Canonical URL of the article, may differ if external, e.g. https://www.giga.de/external.html
  // (required).
  string canonical = 2;

  // AMP URL of the article.
  string amp = 3;
}
```

These URLs are now part of the main `fields` and marked as mandatory like the `headline`. Because of:

- removing non-reusable messages
- "mandatory data shouldn't be in fields" is not an case anymore

stroeer/core/v1/shared.proto
bobaaaaa: Thats true for the actual section tree but not for an navigation. So we need a name what can describe an section tree item and a navigation item

bobaaaaa:
I will update the docs when we have an agreement.

stroeer/core/v1/shared.proto
thisismana: We could pull the fields of `Reference` up here and rename ~Category~ into `Reference`:

```protobuf
message Reference {
  string type = 1;
  string text = 2;
  map<string, string> fields = 3;
 repeated Reference children =4;
}
```

bobaaaaa: haha true :D

bobaaaaa:
```proto
message Reference {
  string type = 1;
  string label = 2;
  string href = 3;
  map<string, string> fields = 4;
  repeated Reference children = 5;
}

message Article {
  // ....
  repeated Reference references = 1;
  Reference section_tree = 2;
}

message Section {
  // ....
  repeated Reference references = 1;
  Reference section_tree = 2;
}
```

stroeer/core/v1/shared.proto
```diff
@@ -6,31 +6,34 @@ option go_package = "github.com/stroeer/tapir/go/stroeer/core/v1;core";
 option java_multiple_files = true;
 option java_package = "de.stroeer.core.v1";
 
+// A Reference represents a link to another entity - for example an `Article`,
```
moritzzimmer, line 9: Maybe we should add the fact, that a reference can be a whole tree structure now in this comment as well

thisismana (approved):

PR #165, opened 2021-02-19
fix: non repeated sections

bobaaaaa:
Oh snap!

Will release `0.13.1` even this is breaking.

### DIFF
--- a/stroeer/page/section/v1/section_page.proto
+++ b/stroeer/page/section/v1/section_page.proto
@@ -11,7 +11,7 @@ option java_package = "de.stroeer.page.section.v1";
 
 // All data to render a section page like the homepage or "/politik/".
 message SectionPage {
-  repeated Section section = 1;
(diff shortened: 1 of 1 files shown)


stroeer/page/section/v1/section_page.proto
```diff
@@ -11,7 +11,7 @@ option java_package = "de.stroeer.page.section.v1";
 
 // All data to render a section page like the homepage or "/politik/".
 message SectionPage {
-  repeated Section section = 1;
+  Section section = 1;
```
shobhitsharma, line 14: 🤘

shobhitsharma (approved):

stroeer/page/section/v1/section_page.proto
thisismana: 🖕section, not 🤘 sections

thisismana (approved):

Issue #57, opened 2020-10-01
Improve build/release process


thisismana:
1. `make` to build all the files locally, for all the supported languages
2. `make release` to create a release and git-tag (ideally auto-incremented by either minor/major/patch)
3. `make test` to locally run tests

stick to one build tool (e.g. no additional scripts such as build_docker.sh)

PR #175, opened 2021-03-02
gh release worklow based on tags

moritzzimmer:
Created a new workflow based on Git tags with a go modules compatible scheme (`vX.X.X`). Version of java and node packages still use the identifier without the `v`.

Examples: a new tag with `v0.0.1` should release `de.stroeer.api.grpc.tapir-v1:0.0.1` for java

After this we should improve automatic releases based on tags in https://github.com/stroeer/tapir/issues/57

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -6,17 +6,7 @@ on:
       - 'v*'
 
 jobs:
-  version:
-    runs-on: ubuntu-latest
-    outputs:
-      tag: ${{ steps.version.outputs.tag }}
-    steps:
-      - name: Version
-        id: version
-        run: echo ::set-output name=tag::${GITHUB_REF/refs\/tags\//}
-
   java:
-    needs: version
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


bobaaaaa (approved):
(nit-pick) For node it is very important that the version set via `npm version` is without `v` prefix. Maybe thats worth a comment

.github/workflows/packages.yml
```diff
@@ -0,0 +1,77 @@
...
+
+jobs:
+  version:
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.version.outputs.tag }}
+    steps:
+      - name: Version
+        id: version
+        run: echo ::set-output name=tag::${GITHUB_REF/refs\/tags\//}
```
thisismana, line 16: You can simplify this line and make the following `$(cut -d "v" -f2 <<< "$TAG")` obsolete if you remove the `v` right in this line.

tested with bash 3.2:

```
$ export GITHUB_REF='refs/tags/v1.0.0'
$ echo ${GITHUB_REF/refs\/tags\/v/}
1.0.0
```

thisismana (changes requested):

moritzzimmer:
> (nit-pick) For node it is very important that the version set via `npm version` is without `v` prefix. Maybe thats worth a comment

@bobaaaaa we'll strip the `v` prefix passed to npm and gradle

.github/workflows/packages.yml
moritzzimmer: @thisismana nice! PR is adapted to this!

thisismana (approved):

PR #221, opened 2021-05-28
StageService (for Schlagzeilen on ArticlePages)

harryurban:


stroeer/content/v1/content_service.proto
```diff
@@ -0,0 +1,58 @@
...
+
+import "stroeer/page/section/v1/section_page.proto";
+
+package stroeer.content.v1;
+
+option go_package = "github.com/stroeer/tapir/go/stroeer/content/v1;content";
+option java_multiple_files = true;
+option java_package = "de.stroeer.content.v1";
+
+service ContentService {
```
thisismana, line 11: As our `ArticlePageService` and `SectionPageService`, this should either be a `FragmentService` (or `Partial(s)Service`) because `Content` is too broad.

Another better fitting name would be `StageService`

thisismana (changes requested):

stroeer/fragment/stage/v1/stage_service.proto
```diff
@@ -0,0 +1,58 @@
...
+option go_package = "github.com/stroeer/tapir/go/stroeer/fragment/stage/v1;stage";
+option java_multiple_files = true;
+option java_package = "de.stroeer.fragment.stage.v1";
+
+service StageService {
+  /**
+    ############################################################################
+    # Description
+    ############################################################################
+    Get single stages by requesting them via agreed tokens, e.g. "schlagzeilen" or "meistgelesen".
```
thisismana, line 16: via well-known tokens?

harryurban: nit-picking?

thisismana (approved):

bobaaaaa (approved):

stroeer/fragment/stage/v1/stage_service.proto
```diff
@@ -0,0 +1,58 @@
...
+    cacheable: no
+  **/
+  rpc GetStages (GetStagesRequest) returns (GetStagesResponse) {
+  }
+}
+
+// Message to request stages independent of SectionPages, for client side rendering or ESIs
+message GetStagesRequest {
+  // an id is a token which is agreed between frontend and backend, for example "schlagzeilen"
+  repeated string ids = 1;
```
moritzzimmer, line 50: Did you consider enumerations here? Feels like this is a finite list of possible Ids which could be selected in a section tool.

stroeer/fragment/stage/v1/stage_service.proto
moritzzimmer, line 16: I find the mix of calling the identifiers `id` and/or `token` in the comments confusing. I'd go with id/identifier

stroeer/fragment/stage/v1/stage_service.proto
```diff
@@ -0,0 +1,58 @@
+syntax = "proto3";
+
+import "stroeer/page/section/v1/section_page.proto";
+
+package stroeer.fragment.stage.v1;
```
moritzzimmer, line 5: Late to the naming party: why `fragment.stage` package? Stages are usually part of a section and fragment seems unrelated for me. Are companions also stages?

Makefile
```diff
@@ -168,3 +168,7 @@ release: clean check check-git-branch ## Releases new version of gRPC source cod
 	git push origin $(NEXT_TAG)
 	git tag -a $(NEXT_GO_TAG) -m "$(NEXT_GO_TAG)"
 	git push origin $(NEXT_GO_TAG)
+
+
+release-local-java:
```
moritzzimmer, line 173: please add `PHONY` and comment for the help target

stroeer/fragment/stage/v1/stage_service.proto
```diff
@@ -0,0 +1,58 @@
...
+
+option go_package = "github.com/stroeer/tapir/go/stroeer/fragment/stage/v1;stage";
```
moritzzimmer, line 7: please add generated go code to the PR

moritzzimmer (changes requested):

stroeer/fragment/stage/v1/stage_service.proto
thisismana: not quite correct, we need this very StageService within the ArticleDetailPage. Its neither `core` nor a `page`, that's why I suggested a `fragment` package. Not sure if we really need the `stage` package.

stroeer/fragment/stage/v1/stage_service.proto
thisismana: This may be a big enum and change quite often, as there are not only ld-lists in it, but also synthethical stuff like `most-read` and such.
We'd have to update the FE, BE and SectionTool every time something will be added :/

stroeer/fragment/stage/v1/stage_service.proto
bobaaaaa: A companion is just a name for a 'area' of content. This content could be (multiple) `stages`, `commercials` and `teaser`

stroeer/fragment/stage/v1/stage_service.proto
harryurban: Das was MaNa sagt.

stroeer/fragment/stage/v1/stage_service.proto
harryurban: done

Makefile
harryurban: done

stroeer/fragment/stage/v1/stage_service.proto
harryurban: done and https://github.com/stroeer/tapir/issues/229

stroeer/fragment/stage/v1/stage_service.proto
harryurban: done

stroeer/fragment/stage/v1/stage_service.proto
moritzzimmer: guess I would skip `stage` in the package path, but fine for me

moritzzimmer (approved):

Issue #229, opened 2021-06-03
automate generation and addition of go code


harryurban:
# WHAT 
The process of adding go code to this repository should be automated.

# WHY
- Manual steps can be forgotten by devs
- Non-go devs should not be obliged to understand the realm of go, especially since go is not a native programming language in our team / project (there are actually very few devs that know and code go)

moritzzimmer:
We could create a separate repo for generated go code (like google does for their APIs) and trigger a remote workflow from here like this: https://blog.marcnuri.com/triggering-github-actions-across-different-repositories

PR #233, opened 2021-06-07
breaking(api): section data rework

bobaaaaa:
These changes represents the current state of our section ui. Some notable changes:

- no recursive stages anymore. only two level stages.
- explicit companion items
- no explicit width for stage items. handled via stage_layout

The docs are wip.

### DIFF
--- a/stroeer/page/section/v1/section_page.proto
+++ b/stroeer/page/section/v1/section_page.proto
@@ -38,18 +38,18 @@ message Section {
   stroeer.core.v1.Reference section_tree = 2;
 }
 
-// A block of teasers - can consist of editorial articles, advertisement and/or stages.
+// A "row" of teasers - can consist of editorial articles, advertisement and/or stages.
 message Stage {
-  // contains stage_title from now on
+  // contains stage_title and sub_navs from now on
   StageConfiguration stage_configuration = 1;
   repeated StageItem stage_items = 2;
-  repeated StageItem companions = 3; // if present do 1/4 + 3/4 width, if not use 4/4 stage_items
+  // optional, always render the column with the Title and SubNav from the StageConfiguration
+  repeated StageItem companions = 3;
 }
 
 // Rendering data for a stage, currently only a fields map.
 // Coordinate stage layouts, especially the `default` layout with UI department.
 message StageConfiguration {
-  // * `stage_title`: represents an optional label to be rendered above a Stage, like "CORONAVIRUS".
   // * `stage_layout`: represents a visual layout, which may affect teasers, colors, and other styles.
   // If entry `stage_layout` is missing, use layout `default`.
   // the stage_layout field can be used to configure full page width commercials (e.g. nativendo, t-online empfiehlt)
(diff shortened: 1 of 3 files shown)


stroeer/page/section/v1/section_page.proto
```diff
@@ -61,49 +67,53 @@ message StageConfiguration {
...
+  // * `stage_layout`: represents a visual layout, which may affect teasers, colors, and other styles.
+  // If entry `stage_layout` is missing, use layout `stream`.
+  // the stage_layout field can be used to configure full page width commercials (e.g. nativendo, t-online empfiehlt)
   map<string, string> fields = 1;
+  repeated TeaserStageItem teaser_stage_items = 2;
+}
+
+// A single teaseable item.
+// alternatives: ClickableItem, LinkedItem, EverythingButAStage
+message TeaserStageItem {
```
bobaaaaa, line 92: Some better naming ideas? Our first draft was `SubStage` + `SubStageItem`

zenderol, line 92: I think the name "Teaser" implies that only an Article Teaser can be used here, I'd prefer "SubStage" naming or something like "ChildStage" or "StageChildItem"

harryurban: I know what you mean, but doesn't a Commercial tease a product?

zenderol: It does, but this is not our Business Domain. If a user clicks on a Teaser, he is not using our Page anymore. A "Teaser" only teases stuff from our domain, in my opinion.

harryurban: // alternatives: ClickableItem, LinkedItem, EverythingButAStage

harryurban (approved):
❤️

harryurban:
friendly reminder: https://github.com/stroeer/tapir/issues/229 (because of go failing)

stroeer/page/section/v1/section_page.proto
```diff
@@ -59,51 +60,49 @@ message StageConfiguration {
   repeated stroeer.core.v1.Reference references = 2;
 }
 
-// A single generic item of a stage.
+// A generic model to represent what can be part of a stage.
 message StageItem {
```
thisismana, line 64: this is virtually the same as `message TeaserStageItem` only difference is an allowed/forbiden field implemented as `oneof`. And again, `Commercial` and `ArticleTeaser` only differ in one field, but is a different message and is using a different mechanism (separate message).
This feels a bit inconsistent to me.

bobaaaaa: Yep you're right. What would your proposal to remove the recursive stages after level 2?

stroeer/page/section/v1/section_page.proto
```diff
@@ -38,18 +38,19 @@ message Section {
   stroeer.core.v1.Reference section_tree = 2;
 }
 
-// A block of teasers - can consist of editorial articles, advertisement and/or stages.
+// A "row" of teasers - can consist of editorial articles, advertisement and/or stages.
 message Stage {
-  repeated StageItem stage_items = 1;
-  StageConfiguration stage_configuration = 2;
+  StageConfiguration stage_configuration = 1;
```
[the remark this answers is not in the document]
bobaaaaa: Unfortunately there're no namespaces in proto. On this file this looks verbose but on consumer usage view this is ok I guess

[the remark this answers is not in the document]
thisismana: i mean the field name, not the message name. I've tested something with "namespaces", check #235 for better enums by just placing them into their corresponding message

stroeer/page/section/v1/section_page.proto
thisismana: I do not like this move, instead we can somehow guarantee this by contract and get rid of this duplication?

stroeer/page/section/v1/section_page.proto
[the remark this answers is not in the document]
thisismana: and then there is the possibility of nesting https://developers.google.com/protocol-buffers/docs/proto3#nested

we can move virtually everything into `message Stage` and see, how this works out

stroeer/page/section/v1/section_page.proto
```diff
@@ -59,51 +60,49 @@ message StageConfiguration {
...
 
-// A single generic item of a stage.
+// A generic model to represent what can be part of a stage.
 message StageItem {
-  // Width instructions for rendering the single item.
-  StageItemWidth stage_item_width = 1;
-  // At most one field will be set at the same time.
-  // The delegate of the StageItem can be another Stage, an editorial article or a Commercial.
+  // Optional flags/hints for customization independent of the stage_item_type
+  map<string, string> fields = 1;
```
thisismana, line 66: maybe reserve some space here because its expected that the `oneof` will grow in the future

PR #235, opened 2021-06-10
breaking: simple enums

thisismana:
enum naming as described here: https://cloud.google.com/apis/design/naming_convention#enum_names


bobaaaaa (approved):
I like 👍

bobaaaaa (approved):

PR #239, opened 2021-06-15
📝 add an example script for tapir node module

jmeischner:


bobaaaaa (approved):
Thx. For the future we will provide more examples as code.

PR #253, opened 2021-07-07
draft: pagination

thisismana:
- allow page based paging for the SectionPageService
- allow  token based paging for the ArticleService
    - either again page based
    - or rpc streaming (experimental)

Resources to consider:

- https://google.aip.dev/158 
> The response must not be a streaming response.

### DIFF
--- a/stroeer/core/v1/core_article_service.proto
+++ b/stroeer/core/v1/core_article_service.proto
@@ -10,10 +10,42 @@ option java_package = "de.stroeer.core.v1";
 
 // Service to manage core articles
 service ArticleService {
-  // Returns a core article by it's id
+  // Returns a core article by it's `id`
   rpc GetArticle (GetArticleRequest) returns (stroeer.core.v1.Article) {}
+
+  // Returns pages of articles by their `home_section`
+  rpc ListArticles (ListArticlesRequest) returns (ListArticlesResponse) {}
+  // also as continuous stream
+  rpc LotsOfArticles (ListArticlesRequest) returns (stream ListArticlesResponse) {}
 }
 
 message GetArticleRequest {
   string id = 1;
+}
+
+message ListArticlesRequest {
+  // mandatory: valid section_path, with leading and trailing slash
+  // e.g. `/sport/`
+  string section_path = 1;
+
+  // The maximum number of books to return. The service may return fewer than
+  // this value.
+  // If unspecified, at most 10 articles will be returned.
+  // The maximum value is 50; values above 1000 will be coerced to 1000.
+  int32 page_size = 2;
+
+  // A page token, received from a previous `ListArticles` call.
+  // Provide this to retrieve the subsequent page.
+  //
+  // When paginating, all other parameters provided to `ListArticles` must match
+  // the call that provided the page token.
+  string page_token = 3;
+}
+
+message ListArticlesResponse {
+  repeated stroeer.core.v1.Article articles = 1;
+
+  // A token that can be sent as `page_token` to retrieve the next page.
+  // If this field is omitted, there are no subsequent pages.
+  string next_page_token = 2;
 }
--- a/stroeer/core/v1/core_article_service.proto
+++ b/stroeer/core/v1/core_article_service.proto
@@ -15,8 +15,6 @@ service ArticleService {
 
   // Returns pages of articles by their `home_section`
   rpc ListArticles (ListArticlesRequest) returns (ListArticlesResponse) {}
-  // also as continuous stream
-  rpc LotsOfArticles (ListArticlesRequest) returns (stream ListArticlesResponse) {}
 }
 
 message GetArticleRequest {
@@ -28,10 +26,10 @@ message ListArticlesRequest {
   // e.g. `/sport/`
   string section_path = 1;
 
-  // The maximum number of books to return. The service may return fewer than
+  // The maximum number of articles to return. The service may return fewer than
   // this value.
   // If unspecified, at most 10 articles will be returned.
-  // The maximum value is 50; values above 1000 will be coerced to 1000.
+  // The maximum value is 100; values above 100 will be coerced to 100.
   int32 page_size = 2;
 
   // A page token, received from a previous `ListArticles` call.
--- a/stroeer/page/section/v1/section_page_service.proto
+++ b/stroeer/page/section/v1/section_page_service.proto
@@ -71,12 +71,25 @@ service SectionPageService {
 // Correct paths have a leading and a trailing slash, like `/nachrichten/unterhaltung/`
 // The homepage has the path `/`.
 message GetSectionPageRequest {
+  // valid section_path, with leading and trailing slash
   string section_path = 1;
+
+  // use to page through sections. If unspecified, it will default to `1`.
+  // Paging is 1-based (1 is the first page, there is no page `0`)
+  //
+  // Due to underlying mechanisms and seo requirements, page-size is fixed at 30
+  // The service may return fewer than this value.
+  int32 page = 2;
 }
 
 // Response message when requesting data for a section page.
 // Responds with `NOT_FOUND` if an unknown path is requested, or the path is incorrect.
 message GetSectionPageResponse {
   stroeer.page.section.v1.SectionPage section_page = 1;
-}
 
+  // if request contained a `page`, this page is confirmed here with the same value
+  int32 page = 2;
+
+  // last page that has content for`section_path`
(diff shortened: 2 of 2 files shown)


stroeer/core/v1/core_article_service.proto
```diff
@@ -10,10 +10,42 @@ option java_package = "de.stroeer.core.v1";
...
 // Service to manage core articles
 service ArticleService {
-  // Returns a core article by it's id
+  // Returns a core article by it's `id`
   rpc GetArticle (GetArticleRequest) returns (stroeer.core.v1.Article) {}
+
+  // Returns pages of articles by their `home_section`
+  rpc ListArticles (ListArticlesRequest) returns (ListArticlesResponse) {}
+  // also as continuous stream
+  rpc LotsOfArticles (ListArticlesRequest) returns (stream ListArticlesResponse) {}
```
thisismana, line 19: this `stream` is going to be interesting o__O

stroeer/core/v1/core_article_service.proto
```diff
@@ -10,10 +10,42 @@ option java_package = "de.stroeer.core.v1";
...
+  // When paginating, all other parameters provided to `ListArticles` must match
+  // the call that provided the page token.
+  string page_token = 3;
+}
+
+message ListArticlesResponse {
+  repeated stroeer.core.v1.Article articles = 1;
+
+  // A token that can be sent as `page_token` to retrieve the next page.
+  // If this field is omitted, there are no subsequent pages.
```
moritzzimmer, line 49:
```suggestion
  // If this field is empty, there are no subsequent pages.
```

stroeer/core/v1/core_article_service.proto
moritzzimmer: Do we want to provide this, yet?

stroeer/core/v1/core_article_service.proto
```diff
@@ -10,10 +10,42 @@ option java_package = "de.stroeer.core.v1";
...
+
+message ListArticlesRequest {
+  // mandatory: valid section_path, with leading and trailing slash
+  // e.g. `/sport/`
+  string section_path = 1;
+
+  // The maximum number of books to return. The service may return fewer than
+  // this value.
+  // If unspecified, at most 10 articles will be returned.
+  // The maximum value is 50; values above 1000 will be coerced to 1000.
```
moritzzimmer, line 34: c&p from AIP ;) We sell articles not books

Shall we use 100 as the max value for each page (~ what one ddb batch request supports) and 30 as default if unspecified?

stroeer/page/section/v1/section_page_service.proto
```diff
@@ -71,12 +71,25 @@ service SectionPageService {
...
 }
 
 // Response message when requesting data for a section page.
 // Responds with `NOT_FOUND` if an unknown path is requested, or the path is incorrect.
 message GetSectionPageResponse {
   stroeer.page.section.v1.SectionPage section_page = 1;
-}
 
+  // if request contained a `page`, this page is confirmed here with the same value
+  int32 page = 2;
```
moritzzimmer, line 91: what's the benefit of mirroring this value in the response?

stroeer/page/section/v1/section_page_service.proto
```diff
@@ -71,12 +71,25 @@ service SectionPageService {
 // Correct paths have a leading and a trailing slash, like `/nachrichten/unterhaltung/`
 // The homepage has the path `/`.
 message GetSectionPageRequest {
+  // valid section_path, with leading and trailing slash
   string section_path = 1;
+
+  // use to page through sections. If unspecified, it will default to `1`.
+  // Paging is 1-based (1 is the first page, there is no page `0`)
+  //
+  // Due to underlying mechanisms and seo requirements, page-size is fixed at 30
```
moritzzimmer, line 80: not sure if this is going to be `30` in all cases (e.g. current regional sections have a page size of `10`). Maybe we document that the page size is fixed in the backend here.

stroeer/page/section/v1/section_page_service.proto
```diff
@@ -71,12 +71,25 @@ service SectionPageService {
...
 // Response message when requesting data for a section page.
 // Responds with `NOT_FOUND` if an unknown path is requested, or the path is incorrect.
 message GetSectionPageResponse {
   stroeer.page.section.v1.SectionPage section_page = 1;
-}
 
+  // if request contained a `page`, this page is confirmed here with the same value
+  int32 page = 2;
+
+  // last page that has content for`section_path`
```
moritzzimmer, line 93: isn't this the 

> Total number of pages in this `section_path`

?

stroeer/core/v1/core_article_service.proto
thisismana: this is official google slang copied from https://google.aip.dev/158

stroeer/core/v1/core_article_service.proto
thisismana: the ddb limits are `16 MB` or `100` items. so yeah, lets allow 100.

stroeer/page/section/v1/section_page_service.proto
thisismana: I thought of two reasons:

1. its kinda render-ready, so you just need to pass the response to your templates.
2. if the user pages beyond the dataset, we can deliver something instead of 404ing? (e.g. pass him the last page if he requests a page that does not exist?)

This is what currently happens on our website `page_9999999` https://www.t-online.de/gesundheit/ernaehrung/id_12289886/page_9999999/ernaehrung-gesundes-essen-ernaehrungsberatung.html

stroeer/page/section/v1/section_page_service.proto
thisismana: **POs hate this trick**: as long as its dividable, we could absolutely do this:

given internal pagination has a `page_size=30`, for the consumer we do the following to fake a page size of 10 by dividing our pages into 3 virtual pages:

1. `total_pages = 3 * total_pages`
2. `page = page / 3`
3. `skip = ((page + 2) % 3) * 10`
4. `articles = ddb.query(page).items().skip(skip).take(10)`

stroeer/core/v1/core_article_service.proto
shobhitsharma: Elasticsearch can handle upto `10k` hits. Exposing this that returns upto `1000` maybe ok. I like the min `10` and max `100` set for the start.

stroeer/page/section/v1/section_page_service.proto
```diff
@@ -71,12 +71,25 @@ service SectionPageService {
...
+  int32 total_pages = 3;
```
shobhitsharma, line 94: `uint32`?

shobhitsharma: ... unless we store huge number 🍡

thisismana: API says no:

> Request messages for collections should define an `int32 page_size` field, allowing users to specify the maximum number of results to return.
> Response messages for collections may provide an int32 total_size field, providing the user with the total number of items in the list.

it is usually not recommended to use unsigned because it does not exist in all languages (java ); )

stroeer/page/section/v1/section_page_service.proto
thisismana: seo-jörg says this is wrong: if the users request exceeds the total_pages, we deliver a 404.

thisismana: gonna remove this

stroeer/core/v1/core_article_service.proto
thisismana: we're not using ES behind the scenes, that's why we go for 100. the user has to progress through the result set, I think this is ok and works as expected.

stroeer/core/v1/core_article_service.proto
bobaaaaa, line 19: I did not find anything about this in the protov3 spec. It would be interesting what is generated for node

thisismana: I'd like to have this as an experimental feature. Maybe we remove this later if its no use or too complicated to implement.
But the idea is appealing to make one request for `/section/` and then a continuous stream of items is delivered.

moritzzimmer: we could also experiement locally and release it when useful

moritzzimmer (approved):

PR #266, opened 2021-08-16
[BREAKING] Stages on ArticlePages and Reference facelift

harryurban:
- extract Stage to page/stage/stage.proto
- add Stages on ArticlePages
- refine Reference.type and Reference.fields.layout meanings


stroeer/core/v1/shared.proto
```diff
@@ -30,6 +32,7 @@ message Reference {
   // * `target`
   // * `rel`
   // * `flag:internal`
+  // * `layout:tag` or `layout:navigation`
```
bobaaaaa, line 35:
```suggestion
  // * `layout`
```

```
fields.get('layout') === 'tag'
```

bobaaaaa (approved):

PR #272, opened 2021-08-30
breaking: rename content type OEMBED to EMBED

bobaaaaa:
This is a more accurate name for this main type. Because we use this type to transport `edge_side_include` or `oembed` element in it.

### Question:

This change is not breaking our stored proto article data and an outdated service. But it is a breaking change for a code base which uses tapir. Would we label this as a breaking change?


moritzzimmer (approved):

moritzzimmer:
> This is a more accurate name for this main type. Because we use this type to transport `edge_side_include` or `oembed` element in it.
> ### Question:
> 
> This change is not breaking our stored proto article data and an outdated service. But it is a breaking change for a code base which uses tapir. Would we label this as a breaking change?

This should be wire compatible for gRPC users (would be breaking for users of a REST API), so should be able to safely release this.

thisismana (approved):

PR #273, opened 2021-08-31
RFC: article.body

bobaaaaa:
# Info

This PR is not ready to merge. It provides just a base for discussions.

## Changes

Added a new type of `body` level. Something like a root body node.

## Motivation

We are adding more and more of these "root body nodes". The frontend is manually extracting most of them and place them somewhere in the UI. For a regular `Body.Node`, the editor has an impact on the position & style. For root body nodes only on styling. The frontend may change the position.

This change reflects the different between a simple HTML `<p />` and something like a `table_of_contents` for all API consumers.

### DIFF
--- a/stroeer/core/v1/article.proto
+++ b/stroeer/core/v1/article.proto
@@ -66,7 +66,7 @@ message Article {
   * ```
   * Hierarchical section tree information of the article (required).
   * See [`Reference`](reference.html)
-  
+
   * Sample:
   * ```javascript
   * {
@@ -139,6 +139,7 @@ message Article {
   * This field will be `empty`/`null` for the teaser representation (e.g. section or overview pages) of the article.
   */
   Body body = 6;
+  repeated BodyPart bodyParts = 66;
 
   /**
   * ## `metadata`
@@ -491,7 +492,7 @@ message Element {
   * ## `type`
   *
   * Type of an element.
-  * 
+  *
   * ```protobuf
   *   enum ElementType {
   *   // Not specified, not supported or yet unknown.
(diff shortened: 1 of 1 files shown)


stroeer/core/v1/article.proto
```diff
@@ -961,6 +962,54 @@ message Body {
   repeated BodyNode children = 1;
 }
 
+// todo: rename to `Body`
+message BodyPart {
+  /*
+  * ### `types`
+  * | type | description |
+  * | -----| ------------|
+  * | `body` | formatted text and inlined elements |
```
bobaaaaa, line 971: `Body.type = 'body'` is a bit weird. This should be the regular body text

stroeer/core/v1/article.proto
```diff
@@ -961,6 +962,54 @@ message Body {
...
+    * | `img` | inline image element, check `elements` |
+    * | `video` | inline video element, check `elements` |
+    * | `gallery` | inline gallery element, check `elements` |
+    * | `oembed` | inline oEmbed element, check `elements` |
+    */
+    string type = 1;
+    string text = 2;
+    map<string, string> fields = 3;
+    repeated BodyPart.Node children = 4;
+    repeated Element elements = 5;
```
bobaaaaa, line 1009: Nothing changed here. See `BodyNode`

stroeer/core/v1/article.proto
```diff
@@ -139,6 +139,7 @@ message Article {
   * This field will be `empty`/`null` for the teaser representation (e.g. section or overview pages) of the article.
   */
   Body body = 6;
+  repeated BodyPart bodyParts = 66;
```
bobaaaaa, line 142: At the end this could be just `repeated Body body = 6;`

stroeer/core/v1/article.proto
```diff
@@ -961,6 +962,54 @@ message Body {
...
+  * | `disclaimer` | formatted text that holds a disclaimer, e.g. used in `/gesundheit/` or `/kaufbertatung/` |
+  * | `infobox` | todo |
+  * | `table_of_contents` | todo |
+  * | `article_sources`| "Quellenapparat", list like structure containing `children[type=article_source]`. See [Samples](#article_sources). |
+  * | `article_source` | A single "Quellenapparat" item usually containing `text` or `a` |
+  */
+  string type = 1;
+  repeated BodyPart.Node children = 2;
+
+  message Node {
```
bobaaaaa, line 981: It's now nested

stroeer/core/v1/article.proto
```diff
@@ -961,6 +962,54 @@ message Body {
...
+  * | type | description |
+  * | -----| ------------|
+  * | `body` | formatted text and inlined elements |
+  * | `disclaimer` | formatted text that holds a disclaimer, e.g. used in `/gesundheit/` or `/kaufbertatung/` |
+  * | `infobox` | todo |
+  * | `table_of_contents` | todo |
+  * | `article_sources`| "Quellenapparat", list like structure containing `children[type=article_source]`. See [Samples](#article_sources). |
+  * | `article_source` | A single "Quellenapparat" item usually containing `text` or `a` |
+  */
+  string type = 1;
```
bobaaaaa, line 978: An `enum` would be perfect. Maybe something before 1.0

stroeer/core/v1/article.proto
```diff
@@ -961,6 +962,54 @@ message Body {
...
+    *
+    * ### Inlined custom `types`
+    * | type | description |
+    * | -----| ------------|
+    * | `img` | inline image element, check `elements` |
+    * | `video` | inline video element, check `elements` |
+    * | `gallery` | inline gallery element, check `elements` |
+    * | `oembed` | inline oEmbed element, check `elements` |
+    */
+    string type = 1;
```
bobaaaaa, line 1005: An `enum` would be perfect here. Maybe something for 1.0

stroeer/core/v1/article.proto
thisismana: we should still keep in mind that there may be paginated articles, so `type` may not be unique.

stroeer/core/v1/article.proto
```diff
@@ -961,6 +962,54 @@ message Body {
...
 
+// todo: rename to `Body`
+message BodyPart {
+  /*
+  * ### `types`
+  * | type | description |
+  * | -----| ------------|
+  * | `body` | formatted text and inlined elements |
+  * | `disclaimer` | formatted text that holds a disclaimer, e.g. used in `/gesundheit/` or `/kaufbertatung/` |
+  * | `infobox` | todo |
```
thisismana, line 973: infobox is an inline thing, so it should not be here?

stroeer/core/v1/article.proto
bobaaaaa: argh good point

stroeer/core/v1/article.proto
bobaaaaa: 👍

stroeer/core/v1/article.proto
thisismana: go for it. I don't see much updates here. And if something changes, it would most probably an addition. Unlikely, that stuff gets removed.

stroeer/core/v1/article.proto
thisismana: I would rather not do it here, It will become messy. This is merely a list of HTML elements, e.g. upcoming `<table>` will add drölfizig new things here like `table,td,th,tr,thead,tbody`

stroeer/core/v1/article.proto
moritzzimmer: @bobaaaaa do you create that enum?

stroeer/core/v1/article.proto
moritzzimmer: is pagination a blocker here? 

from the top of my head: is it possible to include all pages of an article inside the `body` part including some page break identifiers which are then evaluated in the pages api (like remove unecessary parts for the requested page)?

harryurban: if i understand correctly, the mentioned thingies like toc, disclaimer etc are placed conventionally by the frontend. doesn't that make them eligible to be pulled outside of the body?
this could make paging also easier imo.

bobaaaaa:
follow up: #330

PR #279, opened 2021-09-07
feat/keywords: include extracted keywords in core.Article

moritzzimmer:
- include `Keyword` message from new keyword extraction system in `core.Article`
- re-used the existing field number since those keywords are meant to be used in the future (breaking change)
- deprecated old `entities` as a new field

### DIFF
--- a/stroeer/core/v1/article.proto
+++ b/stroeer/core/v1/article.proto
@@ -163,11 +163,11 @@ message Article {
   /**
   * ## `keywords`
   * ```protobuf
-  * repeated string keywords = 9;
+  * repeated Keyword keywords = 9;
   * ```
-  * Extracted entities from the article body like people, places, organizations etc.
+  * Extracted keywords from the article body like persons, locations, organizations etc.
   */
-  repeated string keywords = 9;
+  repeated Keyword keywords = 9;
 
   /**
   * ## `onwards`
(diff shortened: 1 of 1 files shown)


stroeer/core/v1/article.proto
```diff
@@ -181,6 +181,17 @@ message Article {
...
+  /**
+  * ## `entities`
+  * ```protobuf
+  * repeated string entities = 100;
+  * ```
+  * Extracted entities from the article body like people, places, organizations etc.
+  *
+  * `entities` are deprecated and should not be used by clients, use `keywords` instead.
+  */
+  repeated string entities = 100 [deprecated = true];
```
moritzzimmer, line 193: alternatively, we could add something like `source` to the new `Keyword` message and transport old and new keywords/entities with this message

bobaaaaa: If we delete this field entry in the near future, then it's ok. Otherwise, I would remove the `deprecated` flag.

bobaaaaa (approved):
👍

stroeer/core/v1/article.proto
```diff
@@ -1079,3 +1090,34 @@ message BodyNode {
...
+* @FileArticle Keyword
+*/
+
+/**
+* ```protobuf
+* message Keyword {}
+* ```
+* Extracted keywords from the article body like persons, locations, organizations etc.
+*/
+message Keyword {
```
harryurban, line 1104: will this message be used outside of the article context? if not we could define it within the Article message.

stroeer/core/v1/article.proto
harryurban: do we still need the old entities now that we have the new ones from retresco?

stroeer/core/v1/article.proto
moritzzimmer: 🤷  so far, it's only used in the article context

stroeer/core/v1/article.proto
moritzzimmer: yes, legacy systems e.g. for tracking, themepages and content analystics need both for a transistion period

PR #288, opened 2021-09-17
documentation: aligned ArticlePageService error scenarios to implementation

moritzzimmer:


bobaaaaa (approved):

stroeer/page/article/v1/article_page_service.proto
```diff
@@ -37,28 +37,35 @@ service ArticlePageService {
     scenario: not found
     description: article was not found in the datastore
     gRPC status: NOT_FOUND
-    gRPC error payload: google.rpc.ResourceInfo
+    gRPC error payload: none
```
thisismana, line 40: basically correct, this file could be fundoc'ed so it will appear in the documentation (simply add a `@Article` in the header and that's it.

thisismana (approved):

PR #296, opened 2021-09-20
Add PaginationType field for GetSectionPageResponse

harryurban:


stroeer/page/section/v1/section_page_service.proto
```diff
@@ -93,4 +94,17 @@ message GetSectionPageResponse {
 
   // Total number of pages in this `section_path`
   int32 total_pages = 2;
+
+  PaginationType pagination_type = 3;
+}
+
+enum PaginationType {
```
bobaaaaa, line 101: Maybe as a member of `GetSectionPageResponse`?

`GetSectionPageResponse.PaginationType`

harryurban: good idea

PR #341, opened 2022-01-12
Added ContentType `EXTERNAL`

thisismana:
also updated docs to be more google-like, like [this one](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#syntax)


bobaaaaa (approved):
👍

PR #407, opened 2022-07-01
New rpc calls

thisismana:
- added `CoreArticleService/BatchGetArticle`
- added `CurationService/GetCuration`
- added `CurationService/BatchGetCuration`


stroeer/core/v1/core_article_service.proto
```diff
@@ -36,6 +38,7 @@ option java_package = "de.stroeer.core.v1";
 
 service ArticleService {
   rpc GetArticle (GetArticleRequest) returns (stroeer.core.v1.Article) {}
+  rpc BatchGetArticles (BatchArticlesRequest) returns (BatchArticlesResponse) {}
```
moritzzimmer, line 41: `BatchGetArticlesRequest` and `BatchGetArticlesResponse`

stroeer/core/v1/core_article_service.proto
```diff
@@ -58,12 +61,36 @@ service ArticleService {
...
+* # `⚙︎ BatchGetArticle`
+* ```protobuf
+* rpc BatchGetArticles (BatchArticlesRequest) returns (BatchArticlesResponse) {}
+* ```
+*
+* returns multiple [`stroeer.core.v1.Article`](Article.html) for the given `ids`. The
+* ordering of items will the same ordering as the `ids` requested.
+* If an `id` does not exists, it is omitted in the result (no error will be raised).
+*
+* There is a maximum of 100 items that can be queried in one batch.
```
moritzzimmer, line 80: How do we deal with the 16MB limit (implementation detail)?

stroeer/core/v1/core_curation_service.proto
```diff
@@ -0,0 +1,66 @@
+syntax = "proto3";
+
+/**
+ * @FileArticle ⚙︎ CurationService
+ */
+
+package stroeer.core.v1;
```
moritzzimmer, line 7: I would introduce a new top-level package `curation` for those rpc's

stroeer/core/v1/core_curation_service.proto
```diff
@@ -0,0 +1,66 @@
...
+* curation contains. The response may be empty in case the curation does not contain any items.
+*
+* a `NOT_FOUND` status code will indicate the curation `id` does not exist.
+*/
+message GetCurationRequest {
+  int64 id = 1;
+}
+
+message GetCurationResponse {
+  repeated stroeer.core.v1.Article articles = 1;
```
moritzzimmer, line 46: we could add the `update_time` of the curation list here as additional info

stroeer/core/v1/core_curation_service.proto
```diff
@@ -0,0 +1,66 @@
...
+* curations contain. The response may be empty in case the curation does not contain any items.
+* The ordering of items will the same ordering as the `ids` requested.
+*
+*/
+message BatchGetCurationRequest {
+  repeated int64 id = 1;
+}
+
+message BatchGetCurationResponse {
+  repeated GetCurationResponse curations = 1;
```
moritzzimmer, line 65: maybe use a map of `id` to `GetCurationResponse` for client convenience?

stroeer/core/v1/core_curation_service.proto
```diff
@@ -0,0 +1,66 @@
...
+  repeated stroeer.core.v1.Article articles = 1;
+}
+
+/**
+* # `⚙︎ BatchGetCuration`
+* ```protobuf
+*   rpc BatchGetCuration(BatchGetCurationRequest) returns (BatchGetCurationResponse) {}
+* ```
+*
+* Fetch multiple curations by their id and return the [`repeated stroeer.core.v1.Article`](Article.html) those
```
moritzzimmer, line 55: Do we apply publication checks/filtering of the articles on the server side or leave that to the client? We should document this decisision.

moritzzimmer (changes requested):

stroeer/core/v1/core_article_service.proto
thisismana: this is `163 kB` per item ... should be plenty.

... buuuut, If we really get that far, there is little we can do. I can think of one on these:

- reduce the max item count (to 75 or even less to have some margin)
- expose DDB's `UnprocessedKeys` to the consumer and let him handle this issue (also relevant if table was throttled)

stroeer/core/v1/core_curation_service.proto
thisismana: since this service does not reside in `core` any more, I'd say its a production API which only delivers valid stuff.

A concept for _Preview_-like APIs could be beneficial, e.g.:

-  _Contentful CMS_ provides you with different keys for _Prod_ and _Preview_
- we add a field/switch to each gRPC-service that alters the _Mode_

stroeer/core/v1/core_curation_service.proto
thisismana: jep, added all the things (label, time, id) to make it more accessible by consumers

bobaaaaa (approved):

PR #502, opened 2023-04-04
feat: add RelatedArticle source as string

harryherbig:
### WHAT
RelatedArticle: replace field `source` of type enum with field `related_article_source` of type string

### WHY
https://t-online.atlassian.net/browse/BUZZ-3688
https://t-online.atlassian.net/browse/BUZZ-3700

### DIFF
--- a/stroeer/page/article/v1/article_page.proto
+++ b/stroeer/page/article/v1/article_page.proto
@@ -44,9 +44,17 @@ message RelatedArticle {
   // thus not containing any data that is only required on detail pages).
   stroeer.core.v1.Article article = 1;
 
-  // Deprecated source of the related article in form of an enum.
-  reserved 2;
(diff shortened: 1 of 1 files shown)


stroeer/page/article/v1/article_page.proto
```diff
@@ -44,15 +44,9 @@ message RelatedArticle {
...
 
-// Source of the related article.
-enum RelatedArticleSource {
-  // Not specified.
-  RELATED_ARTICLE_SOURCE_UNSPECIFIED = 0;
-
-  // Article was related manually by the editorial department.
-  RELATED_ARTICLE_SOURCE_EDITORIAL = 1;
+  // Source of the related article.
+  string related_article_source = 3;
```
moritzzimmer, line 51: we don't need an intermediate version with both the enum and the new string field? If that's the case, we could also name the field `source`?!

PR #519, opened 2023-11-09
feat: optional legal filtering param in core article service

thisismana:


stroeer/core/v1/core_article_service.proto
```diff
@@ -71,8 +73,8 @@ message GetArticleRequest {
  */
 message BatchGetArticlesRequest {
   repeated int64 ids = 1;
+  Legal legal = 2;
```
moritzzimmer, line 76: do we filter elements only in the implementation or are unpublished articles also removed from batch responses? this should be clear in the documentation and the filter options

stroeer/core/v1/core_article_service.proto
```diff
@@ -115,6 +117,7 @@ message ListArticlesRequest {
   Filters filters = 2;
   int32 page_size = 3;
   string page_token = 4;
+  Legal legal = 5;
```
moritzzimmer, line 120: Thought of including the term `filter` either in the enum name or the variable, but the distinction to `Filters` should be clear. Maybe add the new filter options (elements, unpublished articles) to the existing filters?

stroeer/core/v1/core_article_service.proto
```diff
@@ -256,6 +259,23 @@ message ListArticlesRequest {
   /** @CodeBlockEnd */
 }
 
+/**
+ * | Enum value         | Description                                                 |
+ * |--------------------|-------------------------------------------------------------|
+ * | `TYPE_UNSPECIFIED` | unspecified                                                 |
+ * | `UNFILTERED`       | include all elements regardless of their validity (default) |
```
harryherbig, line 266: this also affects the main articles, not only the elements

stroeer/core/v1/core_article_service.proto
harryherbig: I also like the idea to add the legal stuff to `Filters`

harryherbig (approved):

PR #525, opened 2024-02-08
update required npm version for node-proto

michael-kleemann:
what
set a required minimum npm version for node-proto instead of fixed version 9.
Node 18 LTS now ships with npm 10.
I assume it is unlikely that new node/npm versions will break functionality.

outcome
When updating node/npm the package won't fail to install.
Right now it blocks our pipeline to build our application.
image

### DIFF
--- a/node-proto/package-lock.json
+++ b/node-proto/package-lock.json
@@ -21,7 +21,7 @@
       },
       "engines": {
         "node": "18",
-        "npm": "9"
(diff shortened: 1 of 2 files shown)


bobaaaaa (approved):

michael-kleemann:
@bobaaaaa thanks for merging. Could you also create a release to make it usable?

PR #526, opened 2024-02-12
chore: remove deprecated field `related_articles` from ArticlePage

harryherbig:
# WHAT
chore: remove deprecated field related_articles from ArticlePage

# WHY
this field was already replaced and consumers paper and paper-95 are migrated

### DIFF
--- a/stroeer/page/article/v1/article_page.proto
+++ b/stroeer/page/article/v1/article_page.proto
@@ -15,10 +15,8 @@ message ArticlePage {
   // Unpublished elements will be filtered according to their metadata.
   stroeer.core.v1.Article article = 1;
 
-  // Additional related editorial articles. Articles which are not published
-  // according to their `Metadata` will be filtered.
-  // deprecated in favor `core.Article.relatedArticles`
-  repeated RelatedArticle related_articles = 2 [deprecated = true];
+  // formerly related_articles
+  reserved 2;
 
   // stages for article pages contain companions, links to topic pages, A-Z Modules etc.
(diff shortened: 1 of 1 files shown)


thisismana (approved):

stroeer/core/v1/article.proto
```diff
@@ -70,7 +69,8 @@ message Article {
   Metadata metadata = 7;
   repeated Element elements = 8;
   repeated Keyword keywords = 9;
-  repeated int64 onwards = 10 [deprecated = true];
```
moritzzimmer, line 73: we shouldn't introduce breaking changes in v1 package of core

moritzzimmer (changes requested):

stroeer/page/article/v1/article_page.proto
```diff
@@ -15,10 +15,8 @@ message ArticlePage {
   // Unpublished elements will be filtered according to their metadata.
   stroeer.core.v1.Article article = 1;
 
-  // Additional related editorial articles. Articles which are not published
-  // according to their `Metadata` will be filtered.
-  // deprecated in favor `core.Article.relatedArticles`
-  repeated RelatedArticle related_articles = 2 [deprecated = true];
+  // formerly related_articles
+  reserved 2;
```
moritzzimmer, line 19: guess we don't need this reservation since all usages of this message are transient and we handle it internally?

moritzzimmer (changes requested):

moritzzimmer (approved):

PR #553, opened 2024-07-02
feat: push tapir to Buf schema registry for each new tag

moritzzimmer:
## what

Push tapir on release to the (public) [buf schema registry](https://buf.build/stroeer/tapir) of tapir.

## why

see https://buf.build/docs/bsr/introduction#bsr-goals for details

✅ automatic documentation, we could remove duplications for funcdoc and simplify tools, see https://buf.build/stroeer/tapir
✅ automatic generation of client SDKs for various proto/grpc versions and languages like java, TypeScript, Go, Swift and Python - we could remove our tooling around packaging and hosting SDKs

🟡 might be free (see [plans](https://buf.build/pricing)), according to ` buf beta price` it would cost $34.00/month for Teams plan which doesn't add much value
🟡 we would need to refactor our imports in the client projects to the buf naming conventions, see examples for [java](https://github.com/stroeer/paperboy/pull/1984) and [go](https://github.com/stroeer/go-tapir/pull/35)

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -97,6 +97,8 @@ jobs:
       - uses: actions/checkout@v4
 
       - uses: bufbuild/buf-setup-action@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
 
       - uses: bufbuild/buf-lint-action@v1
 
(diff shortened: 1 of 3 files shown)


.github/workflows/pull-request.yml
```diff
@@ -97,6 +97,8 @@ jobs:
       - uses: actions/checkout@v4
 
       - uses: bufbuild/buf-setup-action@v1
+        with:
```
harryherbig, line 100: warum brauchen wir die secrets beim pull request?

moritzzimmer: rate limiting, see https://github.com/bufbuild/buf-setup-action?tab=readme-ov-file#github-token

harryherbig: "May prevent..." aber ok

PR #556, opened 2024-07-09
chore: node + node-proto

bobaaaaa:
## What

- dependency updates in both packages (everything latest)
- bump node to 20 (node 18 still works)
- re-format

## Why

- Housekeeping
- Security


.github/workflows/packages.yml
```diff
@@ -12,6 +12,8 @@ jobs:
       - uses: actions/checkout@v4
 
       - uses: bufbuild/buf-setup-action@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
bobaaaaa, line 16: @moritzzimmer I added this to all steps. reason:
Bildschirmfoto 2024-07-09 um 11 39 09

PR #612, opened 2024-12-12
feat: add service definition for the `ArticleExtender` for initial questions

harryherbig:
add proposal for grpc service between paperboy and chatbot for initial questions for the article extender


thisismana (approved):

stroeer/page/article/v1/extender_service.proto
bobaaaaa, line 1: `ArticleExtenderService` -> `article_extender_service.proto`?

harryherbig: gerne

PR #619, opened 2025-01-09
feat: add original article id to article_extender_service.proto

Hustenbonbon:
we need the article id, so that the LLM answer does not include the current article

### DIFF
--- a/stroeer/page/article/v1/article_extender_service.proto
+++ b/stroeer/page/article/v1/article_extender_service.proto
@@ -61,7 +61,7 @@ message GetQuestionsResponse {
  */
 message GetAnswerRequest {
   string question = 1;
-  int64 article_id = 1;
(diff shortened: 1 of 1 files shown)


stroeer/page/article/v1/article_extender_service.proto
```diff
@@ -55,11 +55,13 @@ message GetQuestionsResponse {
  * | Field name       | Type                | Description                                       |
  * |------------------|---------------------|---------------------------------------------------|
  * | `question`       | `string`            | [required] The question for the returned answers. |
+ * | `article_id`     | `int64`             | [required] The original article the user read.
```
harryherbig, line 58: table is broken (missing | at the end)

harryherbig (approved):

harryherbig (approved):

PR #622, opened 2025-01-14
feat: tapir-buf

bobaaaaa:
## What

- new node package including tapir generated models + services with buf (`bufbuild-es`) in cjs.
- deprecated `@stroeer/tapir-proto-v1` (docs only)
- README updates
- bumping some dependencies

## Why

- BSR (Buf Schema Registry) auto generates packages with esm only. See https://buf.build/stroeer/tapir/sdks/v0.53.0
- FE currently needs cjs


node-buf/package.json
```diff
@@ -0,0 +1,113 @@
+{
+  "name": "@stroeer/tapir-buf-v1",
+  "version": "0.53.0",
+  "type": "commonjs",
+  "main": "index.js",
+  "module": "index.js",
+  "sideEffects": false,
+  "description": "T-Online API Repo",
+  "engines": {
+    "node": ">=18"
```
Marcel2603, line 10: should we allow only node 22?

bobaaaaa: Node18 is still in LTS and valid. IMO: this is fine

Marcel2603 (approved):

Marcel2603 (approved):

PR #711, opened 2025-08-12
fix(ci): github packages publish permissions

bobaaaaa:
- removed `node-proto` workflow because it's deprecated
- fixed publish permissions for `node` & `node-buf`


moritzzimmer (changes requested):
@bobaaaaa please also remove `node-proto` from the readme and makefile

moritzzimmer (approved):

PR #733, opened 2026-04-10
chore: improve/debug release cmd

thisismana:
the `--notes --target` syntax is wrong and results in each release having the `--target` in the notes. since `--target` defaults to `main`, we simply omit it.

also adding `--fail-on-no-commits` to prevent empty releases that have no commits since the last release.

https://github.com/stroeer/tapir/releases/tag/v0.62.0


bobaaaaa (approved):
cool 👍

